PR #6, opened 2017-05-01
breaking: Introduce client-wide context.Context

jen20:
This commit makes every method to every Triton client take a context in order that the consumer can set deadlines, timeouts and so forth. This also mandates the removal of the RetryableHTTP library in favour of the standard library HTTP client (without upstreaming changes to RetryableHTTP).

This is a breaking change, however it is a vast improvement and worth taking the hit sooner rather than later.

We also clean up various nits - format strings without formatting values and so forth.

### DIFF
--- a/accounts.go
+++ b/accounts.go
@@ -1,11 +1,11 @@
 package triton
 
 import (
+	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"time"
-
-	"fmt"
 
 	"github.com/hashicorp/errwrap"
 )
@@ -40,9 +40,13 @@ type Account struct {
 
 type GetAccountInput struct{}
 
-func (client *AccountsClient) GetAccount(input *GetAccountInput) (*Account, error) {
+func (client *AccountsClient) GetAccount(ctx context.Context, input *GetAccountInput) (*Account, error) {
+	if ctx == nil {
+		panic(nilContext)
+	}
+
 	path := fmt.Sprintf("/%s", client.accountName)
-	respReader, err := client.executeRequest(http.MethodGet, path, nil)
+	respReader, err := client.executeRequest(ctx, http.MethodGet, path, nil)
 	if respReader != nil {
 		defer respReader.Close()
 	}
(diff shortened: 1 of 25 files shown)


fabrics.go
```diff
@@ -26,9 +27,9 @@ type FabricVLAN struct {
 
 type ListFabricVLANsInput struct{}
 
-func (client *FabricsClient) ListFabricVLANs(*ListFabricVLANsInput) ([]*FabricVLAN, error) {
+func (client *FabricsClient) ListFabricVLANs(ctx context.Context, _ *ListFabricVLANsInput) ([]*FabricVLAN, error) {
```
tgross: Just for my own education, and I realize this isn't a change in this PR: do we have these unused parameters (in various places) as placeholders so that we can change the implementations without breaking the API contract? Given that the consumer would need to swap out the empty/placeholder input struct for a populated one, what does this give us? I may just be missing something here.

accounts.go
```diff
@@ -40,9 +40,13 @@ type Account struct {
 
 type GetAccountInput struct{}
 
-func (client *AccountsClient) GetAccount(input *GetAccountInput) (*Account, error) {
+func (client *AccountsClient) GetAccount(ctx context.Context, input *GetAccountInput) (*Account, error) {
+	if ctx == nil {
+		panic(nilContext)
```
tgross: What's special about this particular API endpoint with respect to panicking if the user passes a `nil` Context? Panicking here seems like it would be a surprise to the user and we're not doing it anywhere else. Maybe a default `context.Background` would be more appropriate?

fabrics.go
jen20: There are two reasons for this:

1) As you say, so that we can add in additional options fields without breaking users (if there are optional arguments - for example filters - it would not be necessary to populate them)

2) Uniformity of the API - it's messy to have some things take an options structure and others not, and requires consulting the documentation more often.

accounts.go
jen20: The idiomatic approach to things requiring a context is to panic if it is nil - we should probably roll that change through the entire API though (or delegate it to the client, though that will lead to worse backtraces).

jen20: @sean- any opinions on this bit?

fabrics.go
tgross: Aye, that all seems sensible.

accounts.go
cheapRoc: Since I'm new to Go as well, how are other libraries migrating to this kind of API constraint with `context`? Are they providing defaults at points such as this (is that even good practice)?

cheapRoc:
This is such a perfect time to enforce this API. Glad to see this PR incoming.

jen20:
Ummm, I didn't merge this, not sure how that happened.

jen20:
However I pushed new changes so I'll open a new PR.

accounts.go
jen20: In this case actually I think we should remove it and let the HTTP library panic on nil context instead - @tgross is right.

PR #8, opened 2017-05-04
breaking: Introduce client-wide context.Context

jen20:
This commit makes every method to every Triton client take a context in order that the consumer can set deadlines, timeouts and so forth. This also mandates the removal of the RetryableHTTP library in favour of the standard library HTTP client (without upstreaming changes to RetryableHTTP).

This is a breaking change, however it is a vast improvement and worth taking the hit early on.

This is an updated version of #6, which GitHub automatically closed for an unknown reason.

### DIFF
--- a/accounts.go
+++ b/accounts.go
@@ -1,11 +1,11 @@
 package triton
 
 import (
+	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"time"
-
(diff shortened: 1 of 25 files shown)


jen20:
@tgross @sean- @cheapRoc I think all of the review points have been addressed now.

Issue #10, opened 2017-05-04
Remove dependency on errwrap


jen20:
We are currently using the HashiCorp `errwrap` library - this is unnecessary since we only need a tiny subset of the functionality.

sean-:
Nuking `errwrap`++.

Bumping this issue given the fun we had on this topic this week.

sean-:
After #90 , are we g2g to close this issue out?

stack72:
Nope - still got #91 and then the compute package!

PR #16, opened 2017-05-11
Remove dependency on hashicorp/errwrap

cheapRoc:
This PR addresses #10 by fork lifting the functionality while still removing the dependency. Our primary use case should still be addressed by including the error wrapping functionality internally to `triton-go`.

Looking for feedback before I continue down the route of replacing every other case throughout our API.

/cc @jen20 @sean- @tgross

### DIFF
--- a/client.go
+++ b/client.go
@@ -12,7 +12,6 @@ import (
 	"net/url"
 	"time"
 
-	"github.com/hashicorp/errwrap"
 	"github.com/joyent/triton-go/authentication"
 )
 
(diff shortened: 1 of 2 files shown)


cheapRoc:
Definitely looking for opinions on whether this is behavior we should even continue to include or if normal errors are reasonable enough. I don't have enough experience to know but it would seem to still be of use if we allow `triton-go` to have varying underlying client types (mock HTTP client vs. `net/http`), etc.

jen20:
I think we should discuss this in #go in chat and decide which approach we want to take here - cc @sean-.

cheapRoc:
We still need better error handling and logging but this is ancient and won't port over well.

☠️ ☠️ ☠️

PR #17, opened 2017-05-18
Patch NPE by constructing a default SSHAgentSigner under NewClient

cheapRoc:
This PR addresses an edge case where a user can cause a self inflicted NPE ([here](https://github.com/joyent/triton-go/blob/master/client.go#L113) and [here](https://github.com/joyent/triton-go/blob/master/client.go#L176)) by not constructing a proper client without passing in an `authentication.Signer`. 

I've added a little more error handling and a new default behavior which utilizes `SSHAgentSigner` if an `SDC_KEY_ID` is found in the user's environment. Also, I believe this is how other HTTP clients act that I've used in the past.

### DIFF
--- a/client.go
+++ b/client.go
@@ -19,6 +19,8 @@ import (
 
 const nilContext = "nil context"
 
+var MissingKeyIdError = errors.New("Default SSH agent authentication requires SDC_KEY_ID")
+
 // Client represents a connection to the Triton API.
 type Client struct {
 	client      *http.Client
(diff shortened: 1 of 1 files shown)


cheapRoc:
Also, there's probably some better logic that could be added around those two `c.authorizers[0]` calls but that's probably over kill...

cheapRoc:
@jen20 Makes sense to include those now.

Issue #9, opened 2017-05-04
Port and merge in manta-go


cheapRoc:
Per offline discussion, we'd like to port over [jen20/manta-go](/jen20/manta-go), along with any other go/manta effort, and include within this single Triton SDK. Come up with the product/package namespace at a future date as well.

cheapRoc:
#23 closed this issue.

Issue #2, opened 2017-01-18
License


tgross:
This repo does not have a LICENSE file. Generally Joyent is using MPL2.0.

However, in https://github.com/docker/distribution/issues/2041#issuecomment-273480408 it's been noted that this license is copyleft and this might raise questions about what projects the library can be reused in.

While the MPL2.0 is probably the best license for our main product lines, this discussion has me questioning whether it's the right license for SDKs like `triton-go`. This has not come up previously because Terraform (the sole consumer of `gomanta` and `gotriton` as far as I know) is itself MPL2.0

cc @misterbisson @bcantrill @jen20

tgross:
[Mozilla's own FAQ describes it that way (see Q12)](https://www.mozilla.org/en-US/MPL/2.0/FAQ/) and I've seen it described as "weak copy-left" in places.

jen20:
As far as I'm aware MPLv2 is the closest thing to the CDDL in common (non-ex-Sun) usage today - it's file based copyleft. Given the usage of this library primarily in Terraform and Packer, which are both MPLv2 license it hasn't been a problem, however I'm in favour of relicensing it as MIT or Apache v2.

Issue #1, opened 2016-12-14
Reuse SSH Agent connections where possible


jen20:
Currently an `authentication.SSHAgentSigner` establishes a new connection to the SSH Agent for each request. It would be beneficial to reuse the connection for multiple requests (and likely a cached copy of the key reference after locating it), as per the Node.js version of this.

cheapRoc:
So I take it that this simply requires multiplexing to the SSH Agent via `SSH_AUTH_SOCK`, correct?

jen20:
Rereading this code we do reuse the connection right now - the problem is we do no maintenance of it. If the agent is restarted during the lifetime of the signer we will likely encounter errors. That seems to me to be the problem we should solve for now.

PR #19, opened 2017-06-05
WIP: Reorganization of the project while porting/merging `manta-go`

cheapRoc:
Ref: #9 

While researching [various](https://github.com/docker/distribution) [things](https://github.com/heptio/eventrouter/tree/master/sinks) I'd like to do with the Manta and Go, I decided it was time to crank out the merge of `manta-go`. 

It also made sense to do this in a way that properly supports both branding of [_Triton Compute_](https://www.joyent.com/triton/compute) and [_Triton Object Storage_](https://www.joyent.com/triton/object-storage), because Product Engineering, and also now is the time to do this sort of thing.

This looks like an enormous degree of changes, and it is, but the API is still sound. I'm trying to keep the changes to the surface area of both APIs as small as possible. There are a few breaking changes. I believe I've listed them below.

I definitely want feedback on whether or not this makes sense. This is sort of a pile driver for the project, but hopefully the last.

Also, **WIP**. Couple small errors and a few things left to change.

_Edit_: Removed reviewer notes in favor of a TODO list.

### ClientConfig

The client is brought up through a new `ClientConfig`. I believe this was a merge from `manta-go`.

```go
client, err := triton.NewClient(&triton.ClientConfig{
        endpoint:    endpoint,
        accountName: accountName,
        signers:     sshKeySigner,
})
```

### Compute and Storage

I've moved `Client` itself into a shared package (like `authentication`) which is shared between both the `compute` and `storage` sub-packages. One client shall rule them all.

`triton-go`  becomes `Compute` and `manta-go` becomes `Storage`. Notice `client.Compute` and `client.Storage` variables which mount the sub-package APIs onto `Client`. This was done with as little changes to the underlying APIs as I could.

```go
machines, err := client.Compute.Machines().ListMachines(context.Background(), &compute.ListMachinesInput{})

stateObject, err := client.Storage.GetObject(&storage.GetObjectInput{
        ObjectPath: "tempstate.tfstate",
})
```

### authentication

Both APIs share the `authentication` package and I folded in missing pieces from each.

### testutils

Since `manta-go` wasn't started with many tests, it was easy to just split out `triton-go`'s test behavior into it's own sub-package.

## TODOs

- [x] Remove `storage/client.go`
- [x] Merge `executeRequest` function calls into a single API off `client`
- [x] Deprecate `go-retryablehttp` from the `storage` package
- [x] Nail down new API
- [x] Implement the new API

_EDIT_: I'll push a new branch when we're done discussing here.


cheapRoc:
Forgot to /cc @misterbisson @tgross @sean- @jen20

jen20:
>go-retryablehttp was brought in from manta-go and lives under the storage package, but not for compute. I was planning to remove it from the storage package if there are no objections. I'm unsure of it's value for the old triton-go functionality.

For some additional context here (prior to review) - `triton-go` originally used `retryable-http` too but it was removed a while back. The same change never got ported to `manta-go` however.

tgross:
This looks pretty good. My only big-picture thought is that perhaps accounts and keys should be in their own top-level package and consumed by the others. Under the hood these are managed by CloudAPI, of course, but from the perspective of end-users keys and accounts are presented as if they were their own resources.

cheapRoc:
I just realized that JPC's dashboard lists each resource already... which makes perfect sense to match with our own top level resources. I'll still come up with a mapping table so we can discuss before pulling the trigger.

screen shot 2017-06-06 at 3 34 11 pm

cheapRoc:
From everyone's feedback, and looking over existing APIs, here's a set of mappings from old groups to new top level cloud resources.

tldr; We'll mostly split up existing `triton-go` APIs into `Account`, `Identity`, `Compute`, `Networking`, `Analytics`, and leave `Storage` for all `manta-go` endpoints. I've also swapped the dashboard's `Access` for `Identity`.

| Old Grouping  | New Top Level | API |
|:---------------|:---------------|:----------------|
| Account       | Account       | CloudAPI       |
| Keys          | Account       | CloudAPI       |
| Config        | Account       | CloudAPI       |
| Users         | Identity      | CloudAPI       |
| Roles         | Identity      | CloudAPI       |
| Role Tags     | Identity      | CloudAPI       |
| Policies      | Identity      | CloudAPI       |
| SSH Keys | Identity      | CloudAPI       |
| Datecenters   | Compute       | CloudAPI       |
| Services      | Compute       | CloudAPI       |
| Images        | Compute       | CloudAPI       |
| Packages      | Compute       | CloudAPI       |
| Instances     | Compute       | CloudAPI       |
| Networks      | Networking    | CloudAPI       |
| Fabrics       | Networking    | CloudAPI       |
| Nics          | Networking    | CloudAPI       |
| FirewallRules | Networking    | CloudAPI       |
| Objects       | Storage       | Manta          |
| Directories   | Storage       | Manta          |
| SnapLinks     | Storage       | Manta          |
| Jobs          | Storage       | Manta          |

Here's some examples of Go usage, including the namespace...

```go
instances, err := client.Compute.Instances().ListMachines(ctx, &compute.ListMachinesInput{})

key, err := client.Account.Keys().CreateKey(ctx, &account.CreateKeyInput{})

role, err := client.Identity.Roles().CreateRole(ctx, &identity.CreateRoleInput{})

vlans, err := client.Networking.Fabrics().ListFabricVLANs(ctx, &networking.ListFabricVLANsInput{})
```

Without the namespace...

```go
instances, err := client.Compute.ListMachines(ctx, &compute.ListMachinesInput{})

key, err := client.Account.CreateKey(ctx, &account.CreateKeyInput{})

role, err := client.Identity.CreateRole(ctx, &identity.CreateRoleInput{})

vlans, err := client.Networking.ListFabricVLANs(ctx, &networking.ListFabricVLANsInput{})
```

Since we're still tracking the existing CloudAPI implementation, I'd rather not rename endpoints like `ListMachines()` to `ListInstances()`. In that regard, I feel like the namespaces still hold a purpose too.

Thoughts?

/cc @misterbisson

cheapRoc:
Going to close this PR and wrap up the remaining effort under a new one.

jen20:
:+1: This looks good to me. I think I would go with the second option, now that the namespaces are focused around services rather than focused around resource types.

cc @sean-

cheapRoc:
I agree... I think option two makes much more sense. Specifically because pointless collisions start happening like `client.Account.Accounts().GetAccount()`.

We'll have...

```go
client.Account.GetAccount()
client.Account.ListKeys()
client.Account.UpdateConfig()
```

sean-:
I have a contrarian point of view on this (apologies in advance).  First, I'd suggest reading https://blog.golang.org/package-names as foreshadowing of my feedback.  Iterating on the above example:

```go
client.Account().GetAccount()
client.Account().ListKeys()
client.Account().UpdateConfig()
```

I'd actually pivot and change to something like:

```go
client.Account().Get()         // No Stutter
client.Account().Keys().List() // One verb per package
client.Account().Config().Update()
```

Which looks silly on one hand, but I've conceptually broken it down in my mind into the following pattern:

```go
// Account is a difficult example for "reasons," can we pick a new type to use to
// illustrate this point? :)
client         .Account()  .Get()
serviceProvider.Namespace().Verb()

client         .Account()  .Keys()    .List()
serviceProvider.Namespace().Resource().Verb() // Resource == Noun

client         .Account()  .Config()  .Update()
serviceProvider.Namespace().Resource().Verb()
```

I fought that convention for a while and have come to appreciate the clarity the rigidity provides.  Also, good use of packages allows for future use of [gomvpkg](https://godoc.org/golang.org/x/tools/cmd/gomvpkg) to wholesale move pieces of our API around as time passes.

The gold-standard of interface design for service providers is, IMO, is Google and their Cloud API.  Their [storage interface](https://godoc.org/cloud.google.com/go/storage) is clean and simple.  In particular, I want to call out the way that a method returns an object or object handle that satisfies various reusable interfaces (e.g. the [`ACLHandle`](https://godoc.org/cloud.google.com/go/storage#ACLHandle) interface).  The [`datastore`](https://godoc.org/cloud.google.com/go/datastore#pkg-index) package is probably a better example to pick up style tips from.

```go
c, err := account.NewClient(...) // Constructor for an Account handle
a, err := c.Get() // Return account details
k, err := c.ListKeys()
updatedAccount, err := c.Update(a)
```

Where this gets more interesting is in the networking and compute examples where the taxonomy is pretty clear:

```go
instances, err := client.Compute.Instances().ListMachines(ctx, &compute.ListMachinesInput{})
// becomes:
c, err := compute.NewClient(ctxt, ...)
vms, err :=  c.ListVMs()

key, err := client.Account.Keys().CreateKey(ctx, &account.CreateKeyInput{})
// becomes:
c, err := compute.NewClient(ctxt, ...)
key, err := key.New(...) // this doesn't require the client
keyHandle, err := c.AddKey(key) // The key was generated locally, we didn't ask the service provider to create a key for us

role, err := client.Identity.Roles().CreateRole(ctx, &identity.CreateRoleInput{})
// becomes:
c, err := identity.NewClient(ctxt, ...) // "identity" == "triton/identity" package
role, err := c.NewRole(...)

vlans, err := client.Networking.Fabrics().ListFabricVLANs(ctx, &networking.ListFabricVLANsInput{})
// becomes:
c, err := network.NewClient(ctxt, ...) // "network" == "triton/network"
f, err := c.GetFabric()
vlans, err := f.GetVLANs()
```

One of the innovations that I sometimes end up on the fence over is whether or not we do actually preserve the `struct *Input{}` idioms for input arguments.  There are two arguments for and against that:

Pro `*Input{}`: source-level forward compatibility
Con: This is the compile-time equivalent of passing `map[string]interface{}` as input and creates a loosely coupled interface where we must assume the zero-value to be correct and arguments may not be present.  Input validation is now a runtime concern of the library and not a compile-time concern, which makes the interface harder to use correctly.

On the other hand, compile-time rigidity provides an explicit compile-time headache for all consumers of the library.

In general I favor compile-time enforcement and checking, but because this is a publicly consumed library, it is inappropriate to make use of compile-time constraints precisely because we want to retain forward compatibility with source code.  With a [hat-tip to Google's Protobuf and their acknowledgement that `optional` is the de facto field rule (over `required`)](https://developers.google.com/protocol-buffers/docs/proto#required_warning), preserving source-code forward compatibility and reducing the maintenance cost of integrating our library into applications is more important than either runtime performance or compile-time rigidity.  The result of this API call is going out over a WAN and we are communicating between two loosely coupled system: a user application and us as a services provider.  The rate of innovation and development is decoupled and therefore we need a loose approximation, namely we should use `*Input{}` structs as arguments (which do give us compile-time type safety, just not function signature safety).

Now that I'm howling at the moon, another nit that I also don't like the Ruby-esque style of chaining return values together in order to compose types that you can perform actions on.  This morning I read http://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html and think it does a good job of articulating the other element of Go style that is worth keeping in mind (I'm not saying we suffer from this, just enumerating this point): embrace correct use of `interface`.

cheapRoc:
/cc @sean- 

From Slack it sounds like we're all in agreement on keeping the `*Input{}` structs. They're a construct I'm finding more and more useful for areas of compatibility and I think they'll serve us well.

This line is important though...

> we must assume the zero-value to be correct and arguments may not be present

Can you expand on the case where we might favor defining required parameters within the signature itself as well as compile-time safety of optional parameters? Instead of forcing everything through the input struct itself.

Wouldn't something like `instances.Update(instanceID, &instance.UpdateOpts{})` provide both compile-time and function sig safety?

I've seen other consumer facing APIs that do this and it appears useful given that the `id` in the above case is almost always going to be required.

I'm still parsing through the other samples and links but I feel like this is perfectly aligning where we want to be. Specifically, the usefulness of clearly defining our packages and the resources that we're acting upon ourselves and providing to users for consumption. Defining not just the client interfaces (`client.Machines().CreateMachine()`) but resources they can act upon within their own runtime (`type Machine struct`). I'll look at how the setup of Google's client pairs with the behavior of our Cloud API.

cheapRoc:
Re-opening due to on-going discussion.

sean-:
@cheapRoc Including the primary key in the function signature works.  So does including something like `LookupByN()` where the required argument is tightly coupled to the function name.  The gist of it is, it should be hard to abuse or misuse the interface and incorrect usage should be pushed as close to compile time as possible with a _strong_ eye and awareness toward the future where API and signature could or would change.

If in doubt, put it in the `*Input{}` argument for optional parameters in order to maintain forward source code compatibility with deployed clients because this API spans organizations (said differently for folks with a background in DNS: this is the "`CNAME` pattern equivalent, but for API semantics").

I'd almost go so far as to say, with public API design, the list of allowable attributes should be on a whitelist of permitted values and everything else should go into `*Input{}`.  The initial whitelist includes:

* IDs (account IDs, user IDs, group IDs, etc)

... other suggestions for what should/could be on that whitelist?  😀

cheapRoc:
@sean- Your comment regarding ID definitely speaks to 99% of REST APIs for sure. Besides string UUIDs, I don't believe Joyent’s CloudAPI has any others to whitelist.

> `LookupByN()` where the required argument is tightly coupled to the function name

I enjoy abstractions like this. They also make great helper functions over base functionality within `List()` endpoints. Then your `N()` becomes an abstraction on searches of `Name` or `UUID`, which match to `/users?username=$name`. Too early to tell if those are of any value here though.

----

I like where this conversation is going in terms of style and refactoring but I fear those will add bloat to the original intent of getting the API redefined/organized. We also need tests, lots and lots of tests.

I’ve sat on this and feel good about two proposals. Here are two API renaming options I've got... 

Option #1 https://gist.github.com/cheapRoc/147a04b4a6e59301830d69d796501ffa
Option #2 https://gist.github.com/cheapRoc/1f3fe48a66da52416f2b2e8bba30ad25

I like that every new namespace group has it’s own client which means you can import what you need and ignore other parts of the library. 

While the former is shorter (sacrificing resource/package clarity), the latter has more explicit resources for every call which I tend to favor in a client library. First time consumers will favor readability over clean package reuse. And it's easy to pair with the CloudAPI docs.

Final thoughts before I wrap this up? I’m going with Option #2.

tgross:
Another advantage of option 2 is that the interfaces one has to implement for each of the components is smaller. Ex. if you have a piece of code that needs to manipulate firewalls and want to provide an interface for mocking, the surface area is just 9 methods instead of 19.

cheapRoc:
Closing this for the new one.

PR #26, opened 2017-07-13
images: Add missing parameters to ListImage operation

jen20:
This will need cherry-picking (or otherwise transferring) into the reorganization but solves a problem I have immediately without needing to refactor the entire API. cc @cheapRoc, @sean-


cheapRoc:
I'll add this to my PR.

PR #27, opened 2017-07-19
Align a few API names for datacenters and instances

cheapRoc:
While integrating with Terraform I found a few bugs in naming around datacenters and instances.


tgross:
LGTM.

Issue #25, opened 2017-07-08
Completeness against CloudAPI


cheapRoc:
The following was extracted from the old README. We can track these as a single issue moving forward.

- Accounts
    - [x] GetAccount
    - [x] UpdateAccount
- Keys
    - [x] ListKeys
    - [x] GetKey
    - [x] CreateKey
    - [x] DeleteKey
- Users
    - [x] ListUsers
    - [x] GetUser
    - [x] CreateUser
    - [x] UpdateUser
    - [ ] ChangeUserPassword
    - [x] DeleteUser
- Roles
    - [x] ListRoles
    - [x] GetRole
    - [x] CreateRole
    - [x] UpdateRole
    - [x] DeleteRole
- Role Tags
    - [ ] SetRoleTags
- Policies
    - [x] ListPolicies
    - [x] GetPolicy
    - [x] CreatePolicy
    - [x] UpdatePolicy
    - [x] DeletePolicy
- User SSH Keys
    - [x] ListUserKeys
    - [x] GetUserKey
    - [x] CreateUserKey
    - [x] DeleteUserKey
- Config
    - [x] GetConfig
    - [x] UpdateConfig
- Datacenters
    - [x] ListDatacenters
    - [x] GetDatacenter
- Services
    - [x] ListServices
- Images
    - [x] ListImages
    - [x] GetImage
    - [x] DeleteImage
    - [x] ExportImage
    - [x] CreateImageFromMachine
    - [x] UpdateImage
- Packages
    - [x] ListPackages
    - [x] GetPackage
- Instances
    - [x] ListMachines
    - [x] GetMachine
    - [x] CreateMachine
    - [x] StopMachine
    - [x] StartMachine
    - [x] RebootMachine
    - [x] ResizeMachine
    - [x] RenameMachine
    - [x] EnableMachineFirewall
    - [x] DisableMachineFirewall
    - [x] CreateMachineSnapshot
    - [x] StartMachineFromSnapshot
    - [x] ListMachineSnapshots
    - [x] GetMachineSnapshot
    - [x] DeleteMachineSnapshot
    - [x] UpdateMachineMetadata
    - [x] ListMachineMetadata
    - [x] GetMachineMetadata
    - [x] DeleteMachineMetadata
    - [x] DeleteAllMachineMetadata
    - [x] AddMachineTags
    - [x] ReplaceMachineTags
    - [x] ListMachineTags
    - [x] GetMachineTag
    - [x] DeleteMachineTag
[...]

PR #29, opened 2017-07-24
Complete Instances Metadata and ListInstances APIs

cheapRoc:
This PR cleans up a few features while patching a mistyped endpoint.

- Patch [an existing bug](https://github.com/joyent/triton-go/blob/master/compute/instances.go#L464) where the `UpdateMetadata` endpoint is misusing `/tags` instead of `/metadata`. 
- Complete the metadata API by introducing `GetMetadata`, `ListMetadata`, `DeleteMetadata`, and `DeleteAllMetadata`. 
- Introduce queries for the `ListInstances` endpoint.

Also, ref: #25

### DIFF
--- a/compute/instances.go
+++ b/compute/instances.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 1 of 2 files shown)


cheapRoc:
Actually, let me break that out since its a critical fix. I'll open a PR over there.

cheapRoc:
Rebased on top of recent master so I can vendor properly.

PR #30, opened 2017-07-24
Update storage examples for new api

gwydirsam:
This commit fixes storage examples for the new API and provides a demo file (foo.txt) for those programs that need it.


gwydirsam:
The important change is not to use SDC_KEY_FILE -- I don't see that documented anywhere.

examples/compute/instances.go
```diff
@@ -15,16 +14,10 @@ import (
 func main() {
 	keyID := os.Getenv("SDC_KEY_ID")
 	accountName := os.Getenv("SDC_ACCOUNT")
-	keyPath := os.Getenv("SDC_KEY_FILE")
```
cheapRoc: 👍 

I'm totally on-board with removing `SDC_KEY_FILE` as it's not idiomatic. It was simply an environment variable I was using to manually force the path for `authentication.NewPrivateKeySigner`. `authentication.NewSSHAgentSigner` is much better here since its less config.

examples/storage/createjob.go
```diff
@@ -15,26 +14,20 @@ import (
 )
 
 func main() {
-	keyID := os.Getenv("SDC_KEY_ID")
-	accountName := os.Getenv("SDC_ACCOUNT")
-	keyPath := os.Getenv("SDC_KEY_FILE")
+	keyID := os.Getenv("MANTA_KEY_ID")
+	accountName := "tritongo"
```
cheapRoc: Let's use `os.Getenv("MANTA_USER")` as `accountName` here so it works out of the box.

examples/storage/createjob.go
```diff
@@ -103,7 +95,7 @@ func main() {
 		log.Fatalf("ListJobs: %s", err)
 	}
 
-	fmt.Printf("Number of jobs: %d\n", jobs.ResultSetSize)
+	fmt.Printf("Result set size: %d\n", jobs.ResultSetSize)
```
cheapRoc: **REALLY** want this output to display the results of the Job. I'm not sure if the library is broken or what. I haven't had the time to explore how results should properly return from Manta through the `GetOutput()` func. Would be awesome if that's something we can do here but I'm ok with doing it later.

examples/storage/getobject.go
```diff
@@ -13,47 +13,55 @@ import (
 )
 
 func main() {
-	keyID := os.Getenv("SDC_KEY_ID")
-	accountName := os.Getenv("SDC_ACCOUNT")
-	keyPath := os.Getenv("SDC_KEY_FILE")
+	keyID := os.Getenv("MANTA_KEY_ID")
+	accountName := "tritongo"
```
cheapRoc: `os.Getenv("MANTA_USER")` as `accountName`.

examples/storage/createjob.go
cheapRoc: You can ignore this, I captured the chunked response in my proxy so I can create a separate issue for it.

cheapRoc:
I'll incorporate much of this PR into a rewrite of the examples.

PR #32, opened 2017-07-28
instances: Patch CNS processing during instance creation

cheapRoc:
CNS was not respected when creating a new instance. This addresses that concern, adds a thorough acceptance test for instance creation, and includes an example for others.

### DIFF
--- a/compute/instances.go
+++ b/compute/instances.go
@@ -28,8 +28,8 @@ const (
 // values are embedded within a Instance's Tags attribute, however they are
 // exposed to the caller as their native types.
 type InstanceCNS struct {
-	Disable    *bool
-	ReversePTR *string
+	Disable    bool
+	ReversePTR string
 	Services   []string
 }
 
@@ -406,6 +406,18 @@ func (c *InstancesClient) Rename(ctx context.Context, input *RenameInstanceInput
 type ReplaceTagsInput struct {
 	ID   string
 	Tags map[string]string
+	CNS  InstanceCNS
+}
+
+// toAPI() merges both Tags and CNS tags into the same map of strings used as
+// the body of our ReplaceTags request.
+func (input ReplaceTagsInput) toAPI() map[string]interface{} {
+	result := map[string]interface{}{}
+	for key, value := range input.Tags {
+		result[key] = value
+	}
+	input.CNS.toTags(result)
+	return result
 }
 
 func (c *InstancesClient) ReplaceTags(ctx context.Context, input *ReplaceTagsInput) error {
@@ -413,7 +425,7 @@ func (c *InstancesClient) ReplaceTags(ctx context.Context, input *ReplaceTagsInp
 	reqInputs := client.RequestInput{
 		Method: http.MethodPut,
 		Path:   path,
-		Body:   input.Tags,
+		Body:   input.toAPI(),
 	}
 	respReader, err := c.client.ExecuteRequest(ctx, reqInputs)
 	if respReader != nil {
@@ -945,10 +957,10 @@ func tagsExtractMeta(tags map[string]interface{}) (InstanceCNS, map[string]inter
 			switch k {
 			case CNSTagDisable:
 				b := raw.(bool)
-				nativeCNS.Disable = &b
+				nativeCNS.Disable = b
 			case CNSTagReversePTR:
 				s := raw.(string)
-				nativeCNS.ReversePTR = &s
+				nativeCNS.ReversePTR = s
 			case CNSTagServices:
 				nativeCNS.Services = strings.Split(raw.(string), ",")
 			default:
@@ -972,16 +984,14 @@ func (api *_Instance) toNative() (*Instance, error) {
 
 // toTags() injects its state information into a Tags map suitable for use to
 // submit an API call to the vmapi machine endpoint
-func (mcns *InstanceCNS) toTags(m map[string]interface{}) {
-	if mcns.Disable != nil {
-		m[CNSTagDisable] = fmt.Sprintf("%t", *mcns.Disable)
+func (cns *InstanceCNS) toTags(m map[string]interface{}) {
+	if cns.Disable {
+		m[CNSTagDisable] = cns.Disable
 	}
-
-	if mcns.ReversePTR != nil {
-		m[CNSTagReversePTR] = &mcns.ReversePTR
+	if cns.ReversePTR != "" {
+		m[CNSTagReversePTR] = cns.ReversePTR
 	}
-
-	if len(mcns.Services) > 0 {
-		m[CNSTagServices] = strings.Join(mcns.Services, ",")
+	if len(cns.Services) > 0 {
+		m[CNSTagServices] = strings.Join(cns.Services, ",")
 	}
 }
--- a/compute/instances.go
+++ b/compute/instances.go
@@ -254,18 +254,12 @@ func (input *CreateInstanceInput) toAPI() map[string]interface{} {
 		result[fmt.Sprintf("tag.%s", key)] = value
 	}
 
-	// Deliberately clobber any user-specified Tags with the attributes from the
-	// CNS struct.
-	//
-	// NOTE(justinwr): CNSTagServices needs to be tagged so that our instance
-	// can be provisioned with CNS. No other CNS tags will be handled at this
-	// time.
+	// NOTE(justinwr): CNSTagServices needs to be a tag if available. No other
+	// CNS tags will be handled at this time.
 	input.CNS.toTags(result)
-	for key, value := range result {
-		if key == CNSTagServices {
-			result[fmt.Sprintf("tag.%s", key)] = value
-			delete(result, key)
-		}
+	if val, found := result[CNSTagServices]; found {
+		result["tag."+CNSTagServices] = val
+		delete(result, CNSTagServices)
 	}
 
 	for key, value := range input.Metadata {
@@ -409,8 +403,8 @@ type ReplaceTagsInput struct {
 	CNS  InstanceCNS
 }
 
-// toAPI() merges both Tags and CNS tags into the same map of strings used as
-// the body of our ReplaceTags request.
+// toAPI is used to join Tags and CNS tags into the same JSON object before
+// sending an API request to the API gateway.
 func (input ReplaceTagsInput) toAPI() map[string]interface{} {
 	result := map[string]interface{}{}
 	for key, value := range input.Tags {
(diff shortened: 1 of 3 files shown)


compute/instances.go
```diff
@@ -963,8 +974,7 @@ func (api *_Instance) toNative() (*Instance, error) {
 // submit an API call to the vmapi machine endpoint
 func (mcns *InstanceCNS) toTags(m map[string]interface{}) {
 	if mcns.Disable != nil {
-		s := fmt.Sprintf("%t", mcns.Disable)
-		m[CNSTagDisable] = &s
+		m[CNSTagDisable] = fmt.Sprintf("%t", *mcns.Disable)
```
jen20: This should likely use `strconv.FormatBool` from the standard library instead of `fmt.Sprintf`

cheapRoc:
Actually, looks like there are some changes for supporting CNS updates through the ReplaceTags endpoint. I'll include them here as well.

compute/instances.go
cheapRoc: Fair enough, I think I'm going to remove the pointers here as well. I'm not sure why they're used by the `InstanceCNS` struct values.

compute/instances.go
```diff
@@ -28,8 +29,8 @@ const (
 // values are embedded within a Instance's Tags attribute, however they are
 // exposed to the caller as their native types.
 type InstanceCNS struct {
-	Disable    *bool
-	ReversePTR *string
+	Disable    bool
+	ReversePTR string
```
cheapRoc: We should be fine changing these semantics considering this struct was renamed under the new API.

cheapRoc:
Nevermind... found another last minute change.

compute/instances.go
```diff
@@ -961,17 +984,14 @@ func (api *_Instance) toNative() (*Instance, error) {
 
 // toTags() injects its state information into a Tags map suitable for use to
 // submit an API call to the vmapi machine endpoint
-func (mcns *InstanceCNS) toTags(m map[string]interface{}) {
-	if mcns.Disable != nil {
-		s := fmt.Sprintf("%t", mcns.Disable)
-		m[CNSTagDisable] = &s
+func (cns *InstanceCNS) toTags(m map[string]interface{}) {
+	if cns.Disable {
+		m[CNSTagDisable] = cns.Disable
```
cheapRoc: Just confirmed that this actually does need to be a boolean and posted as such through the tags API (as JSON).

cheapRoc:
Ok, this is ready for review again.

compute/instances.go
sean-: A `bool` or a "string masquerading as a `bool`" ?  Regardless, this may be worth calling out as a comment in the code:

> // NOTE(justinwr@): The API requires the CNSTagDisable attribute to be a boolean, not a bool string.

That comment probably needs to chase into a `Validator()` function, actually.

compute/instances.go
```diff
@@ -961,17 +984,14 @@ func (api *_Instance) toNative() (*Instance, error) {
...
-	if mcns.ReversePTR != nil {
-		m[CNSTagReversePTR] = &mcns.ReversePTR
+	if cns.ReversePTR != "" {
+		m[CNSTagReversePTR] = cns.ReversePTR
 	}
-
-	if len(mcns.Services) > 0 {
-		m[CNSTagServices] = strings.Join(mcns.Services, ",")
+	if len(cns.Services) > 0 {
+		m[CNSTagServices] = strings.Join(cns.Services, ",")
```
sean-: In order to get a stable comparison, in both the *Read call and here, do we need to sort `cns.Services`?  I don't know but seem to recall there wasn't a sorting guarantee here (I also don't think we should depend on it if there is one currently).

compute/instances_test.go
```diff
@@ -30,6 +35,202 @@ func getAnyInstanceID(t *testing.T, client *compute.ComputeClient) (string, erro
 	return "", errors.New("no machines configured")
 }
 
+func RandInt() int {
+	reseed()
+	return rand.New(rand.NewSource(time.Now().UnixNano())).Int()
```
sean-: I'm going to stump the seeding library here, either for explicit use or to crib an example from: https://github.com/sean-/seed

examples/compute/create_instance.go
```diff
@@ -0,0 +1,83 @@
...
+	"time"
+
+	triton "github.com/joyent/triton-go"
+	"github.com/joyent/triton-go/authentication"
+	"github.com/joyent/triton-go/compute"
+)
+
+func main() {
+	keyID := os.Getenv("SDC_KEY_ID")
+	accountName := os.Getenv("SDC_ACCOUNT")
```
sean-: It may be worth using `os.LookupEnv` here instead.  This is just an example, but these values can't be empty strings without blowing up the downstream callers.

examples/compute/create_instance.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		log.Fatalf("NewClient(): %s", err)
+	}
+
+	// Create a new instance using our input attributes...
+	// https://github.com/joyent/triton-go/blob/master/compute/instances.go#L206
+	createInput := &compute.CreateInstanceInput{
+		Name:     "go-test1",
+		Package:  "g4-highcpu-512M",
+		Image:    "1f32508c-e6e9-11e6-bc05-8fea9e979940",
```
sean-: What datacenter is this instance out of?  It may be worth noting the image ID is tied to a DC atm.

examples/compute/create_instance.go
```diff
@@ -0,0 +1,83 @@
...
+				state <- instance
+			} else {
+				fmt.Print(".")
+			}
+		}
+	}(created.ID, c)
+
+	select {
+	case instance := <-state:
+		fmt.Println("\nDuration:", time.Since(startTime).String())
```
sean-:
```go
fmt.Printf("\nDuration: %s\n", time.Since(startTime))
```

examples/compute/create_instance.go
```diff
@@ -0,0 +1,83 @@
...
+			Services: []string{"frontend", "web"},
+		},
+	}
+	startTime := time.Now()
+	created, err := c.Instances().Create(context.Background(), createInput)
+	if err != nil {
+		log.Fatalf("Create(): %v", err)
+	}
+
+	// Wait for provisioning to complete...
```
sean-: This should probably be a helper function that we write and expose to consumers because most API consumers are going to want a synchronous API and semantics (including Terraform), not async.

compute/instances_test.go
tgross: Given we actually _need_ cryptographically secure random number generation in the `authentication` package (ref https://github.com/joyent/triton-go/blob/5f7872646922b73632f434afa5e45fcd18aa9d5b/authentication/private_key_signer.go#L77) which is imported into the client, which is in turn imported everywhere, maybe we should put any seeding there?

compute/instances.go
```diff
@@ -395,14 +406,26 @@ func (c *InstancesClient) Rename(ctx context.Context, input *RenameInstanceInput
 type ReplaceTagsInput struct {
 	ID   string
 	Tags map[string]string
+	CNS  InstanceCNS
+}
+
+// toAPI() merges both Tags and CNS tags into the same map of strings used as
+// the body of our ReplaceTags request.
```
cheapRoc: ✔️

compute/instances.go
```diff
@@ -395,14 +406,26 @@ func (c *InstancesClient) Rename(ctx context.Context, input *RenameInstanceInput
...
+func (input ReplaceTagsInput) toAPI() map[string]interface{} {
+	result := map[string]interface{}{}
+	for key, value := range input.Tags {
+		result[key] = value
```
cheapRoc: I think we're alright since we're liberal with what can be a tag and we're also relying on the JSON encoder to marshal types going into CloudAPI. But I can see how the use of `interface{}` would normally allow other types through in cases like this.

compute/instances.go
```diff
@@ -249,13 +249,24 @@ func (input *CreateInstanceInput) toAPI() map[string]interface{} {
...
 	// Deliberately clobber any user-specified Tags with the attributes from the
 	// CNS struct.
+	//
+	// NOTE(justinwr): CNSTagServices needs to be tagged so that our instance
+	// can be provisioned with CNS. No other CNS tags will be handled at this
+	// time.
 	input.CNS.toTags(result)
+	for key, value := range result {
+		if key == CNSTagServices {
+			result[fmt.Sprintf("tag.%s", key)] = value
```
cheapRoc: I can definitely clean this up a bit and rely more on `CNSTagServices`. That is the only tag we need to prepend with `tag.` here, anything else slips through.

compute/instances.go
cheapRoc: A `bool` `bool` since the JSON encoder will keep it a boolean. Doesn't look like we had any other examples before this.

I'll add the comment here as well.

examples/compute/create_instance.go
cheapRoc: This was sloppy. I had written out the example for someone else and refactored into the test. I just need to pull the image ID dynamically as I did there.

examples/compute/create_instance.go
cheapRoc: Yes, I agree. I'll push a PR separately for that so we can define the API.

examples/compute/create_instance.go
cheapRoc: ✔️

examples/compute/create_instance.go
cheapRoc: ✔️

compute/instances.go
jen20: Agreed.

compute/instances.go
jen20: I'd say this is something Terraform should handle rather than the library since that is where the sorted version is required.

examples/compute/create_instance.go
jen20: Seems reasonable. We likely want to define a uniform way of dealing with this for as many resources as possible in order to reduce the amount of API internals which need to be understood to consume this.

compute/instances_test.go
cheapRoc: 👍 I agree and I'll fit this into a secondary PR.

cheapRoc: Ref: #33  ✔️

examples/compute/create_instance.go
cheapRoc: Ref: #34 ✔️

compute/instances.go
cheapRoc: ✔️

cheapRoc:
Think we're good to merge here.

jen20:
Looks good to me!

PR #36, opened 2017-08-10
error: Change from TritonError to client.ClientError

stack72:
Paired with @jen20 to diagnose that the wrong error casting was being
done - therefore, compute.IsSpecificError was failing


jen20:
Good catch, this should fix the error predicate methods. cc @cheapRoc.

PR #38, opened 2017-08-10
Use proper error types in client.DecodeError{}

sean-:
The original bug stems from `client.DecodeError()` returning a `client.ClientError{}` not a `client.TritonError{}`.  Chase relocating `compute.TritonError` into `client.TritonError` which is the leaf in the package import graph.

The incorrect error type was discovered while using the Terraform Triton provider.

Discussed with: @cheapRoc and @stack72

### DIFF
--- a/account/account_test.go
+++ b/account/account_test.go
@@ -5,7 +5,6 @@ import (
 	"testing"
 
 	triton "github.com/joyent/triton-go"
-	"github.com/joyent/triton-go/account"
 	"github.com/joyent/triton-go/testutils"
 )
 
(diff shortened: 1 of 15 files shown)


cheapRoc:
Apparently our pair for this PR was a bit overzealous and also broke the tests, the reason of which we can leave to a secondary Issue (wtf packages?).

I've opened #39 with just the `TritonError` fixes applied.

cheapRoc:
The way I'm running tests is all wrong and forever will I know better. :)

Issue #35, opened 2017-08-04
MANTA_NO_AUTH=1 environment variable is not honored


dekobon:
When running within a Manta zone, `MANTA_NO_AUTH=1` is set. This allows for an application to connect to Manta using the running account without doing HTTP signature authentication.

You can't run golang apps directly in Manta zones / jobs becausetriton-go doesn't disable HTTP signatures when this environment variable is present and enabled.

cheapRoc:
Thanks Elijah. Sounds like I can patch this relatively easy since this is just a shut off valve around our inclusion of HTTP signature authentication to outgoing requests.

/cc @misterbisson

cheapRoc:
I have a bug fix in the works but I'm hitting an issue with `crypto/x509` not finding the default location for SSL root certificates under SmartOS. Still researching...

dekobon:
Is there a way to disable the who crypto module entirely if you are running with `MANTA_NO_AUTH` enabled?

cheapRoc:
I killed the open PR because there are a couple small details I missed the first go.

- We're looking for `MANTA_NO_AUTH=true`
- We should skip initialization of an SSH key signer at client config time. There are log warnings against this today and `MANTA_NO_AUTH` should ignore them.
- Given that `MANTA_URL=http://localhost:80/` is set within a Manta zone, this probably means consumers of our SDK shouldn't have to manually set `s.Client.InsecureSkipTLSVerify()`. This is a little sloppy today and I feel like it can be safely handled for the consumer based on whatever API endpoint/URL they pass in.

I should note that I feel like the Go SDK should not utilize default environment variables to self configure the clients. How those values arrive when configuring a client is up to the consumer and we don't enforce any naming conventions like `MANTA_USER` or `MANTA_ACCOUNT` (except in our examples).

With all of that said, here's an example I'm working with...

```go
package main

import (
	"context"
	"fmt"
	"log"
	"os"

	triton "github.com/joyent/triton-go"
	"github.com/joyent/triton-go/storage"
)

func main() {
	mantaUser, foundUser := os.LookupEnv("MANTA_USER")
	if !foundUser {
		log.Fatal("Couldn't find \"MANTA_USER\" in your environment")
	}

	mantaURL, foundURL := os.LookupEnv("MANTA_URL")
	if !foundURL {
		log.Fatal("Couldn't find \"MANTA_URL\" in your environment")
	}

	s, err := storage.NewClient(&triton.ClientConfig{
[...]

PR #41, opened 2017-08-15
client: Support MANTA_NO_AUTH within Manta zones

cheapRoc:
Ref: joyent/triton-go#35

This adds support for turning off Joyent's SSH based HTTP signature
authentication when running within a Manta zone. The Manta zone detection
utilizes a known Manta image feature which injects an environment variable
called `MANTA_NO_AUTH`. The only other present requirement I've found for
running `triton-go` within Manta is to make sure root TLS certificates are
installed on the SmartOS system at `/etc/ssl/certs/ca-certificates.crt`.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -180,6 +180,9 @@ func (c *Client) ExecuteRequestURIParams(ctx context.Context, inputs RequestInpu
 
 	dateHeader := time.Now().UTC().Format(time.RFC1123)
 	req.Header.Set("date", dateHeader)
+	req.Header.Set("Accept", "application/json")
+	req.Header.Set("Accept-Version", "8")
+	req.Header.Set("User-Agent", "triton-go Client API")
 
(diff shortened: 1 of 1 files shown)


client/client.go
```diff
@@ -180,6 +180,9 @@ func (c *Client) ExecuteRequestURIParams(ctx context.Context, inputs RequestInpu
 
 	dateHeader := time.Now().UTC().Format(time.RFC1123)
 	req.Header.Set("date", dateHeader)
+	req.Header.Set("Accept", "application/json")
```
dekobon: I suggest normalizing the capitalization for headers so that all headers use the same form.

cheapRoc:
Let's wait to have this merged in. I realized we can probably remove the need to pass in a key during configuration since authentication is ignored entirely under this case.

cheapRoc:
I'm actually having issues with this now. Going to close this for the time being.

PR #42, opened 2017-09-11
authentication: Support SHA256 Key IDs

jen20:
This commit adds support for keys in the form SHA256:fingerprint in addition to MD5:fingerprint and fingerprints with no additional information regarding their contents. This is useful since the Triton documentation suggests using `ssh-keygen -l -f ~/.ssh/id_rsa` which by default returns a SHA256 formatted fingerprint in newer versions of OpenSSH.

cc @cheapRoc @stack72 @misterbisson

### DIFF
--- a/authentication/ssh_agent_signer.go
+++ b/authentication/ssh_agent_signer.go
@@ -2,6 +2,8 @@ package authentication
 
 import (
 	"crypto/md5"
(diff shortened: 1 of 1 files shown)


stack72:
Worked with @jen20 on this via zoom and it's working as expected :)

PR #44, opened 2017-09-20
Patch where Triton might return a null response

cheapRoc:
Easy one that came up in the Terraform provider.

Tracking at terraform-providers/terraform-provider-triton#37


stack72:
LGTM! Nice catch

PR #45, opened 2017-09-25
Fix Snaplink Headers

stack72:
When making a snaplink via triton-go, the presence of:

```
headers.Set("Content-Type", "application/json; type=link")
```

was causing the snaplink to fail with the error:

```
{
	"code": "SourceObjectNotFound",
	"message": "/stor/stack72-test3/assets.zip was not found"
}
```

By examining the manta cli tools, I was able to find that the cli doesn't need the presence of a Content-Type header. Therefore, removing the Content-Type header, and we get the correct behaviour - which in this case is a 204 response


cheapRoc:
👍 LGTM

PR #54, opened 2017-11-06
Fix endpoint for compute.ListMachineRules

stack72:
This wasn't as per the documentation https://apidocs.joyent.com/cloudapi/#ListMachineFirewallRules

@cheapRoc FYI ;)


stack72:
Nearly all of the Compute package needs tests - I can add a ticket to make sure we come back and fix these up

PR #56, opened 2017-11-07
Add support for CreateSnapshot and StartMachineFromSnapshot endpoints

stack72:
Add the missing endpoints for snapshot management

### DIFF
--- a/compute/snapshots.go
+++ b/compute/snapshots.go
@@ -99,3 +99,59 @@ func (c *SnapshotsClient) Delete(ctx context.Context, input *DeleteSnapshotInput
 
 	return nil
 }
+
+type StartMachineFromSnapshotInput struct {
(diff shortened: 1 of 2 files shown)


compute/snapshots.go
```diff
@@ -99,3 +99,59 @@ func (c *SnapshotsClient) Delete(ctx context.Context, input *DeleteSnapshotInput
 
 	return nil
 }
+
+type StartMachineFromSnapshotInput struct {
+	MachineID string
+	Name      string
+}
+
+func (c *SnapshotsClient) StartMachineFromSnapshot(ctx context.Context, input *StartMachineFromSnapshotInput) error {
```
cheapRoc: From chat it sounds like we'll change this to `compute.Snapshots().StartMachine()`.

PR #59, opened 2017-11-13
Add support for ListRulesMachines Endpoint

stack72:
@cheapRoc

### DIFF
--- a/network/firewall.go
+++ b/network/firewall.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"net/http"
 
+	"time"
+
 	"github.com/hashicorp/errwrap"
 	"github.com/joyent/triton-go/client"
 )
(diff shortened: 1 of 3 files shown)


network/firewall_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+func listRuleMachinesSuccess(req *http.Request) (*http.Response, error) {
+	header := http.Header{}
+	header.Add("Content-Type", "application/json")
+
+	body := strings.NewReader(`[
+	{
+    "id": "b6979942-7d5d-4fe6-a2ec-b812e950625a",
+    "name": "test",
+    "type": "smartmachine",
```
cheapRoc: Is there anything different between a Docker container or native machine? Are we testing that?

network/firewall.go
```diff
@@ -248,3 +250,52 @@ func (c *FirewallClient) ListMachineRules(ctx context.Context, input *ListMachin
...
+	Brand           string                 `json:"brand"`
+	State           string                 `json:"state"`
+	Image           string                 `json:"image"`
+	Memory          int                    `json:"memory"`
+	Disk            int                    `json:"disk"`
+	Metadata        map[string]string      `json:"metadata"`
+	Tags            map[string]interface{} `json:"tags"`
+	Created         time.Time              `json:"created"`
+	Updated         time.Time              `json:"updated"`
+	Docker          bool                   `json:"docker"`
```
cheapRoc: Minor nit but I don't see this exercised through the new API's unit tests. See note below.

network/firewall_test.go
stack72: We are not testing that - the API itself should take care of that - there are no docs about what is supported here. We will just return a list of machines as expected

cheapRoc: 👍 Pretty much all I wanted to know.

PR #63, opened 2017-12-01
Enable HTTP Keepalives for the Triton client.

sean-:
The new defaults are:

1) Allow HTTP KeepAlives
2) 15sec idle timeout per connection
3) 2x connections per host
4) Default max of 10x connections total for the HTTP client

The idle timeout could be made more aggressive if necessary (e.g. 5s)
but 15s seems like a reasonable default.


cheapRoc:
That looks like it'll do it. 👍

PR #68, opened 2017-12-11
Use path.Join() and path.Clean() when constructing object paths.

sean-:
Use of `fmt.Sprintf()` was bothering my sense of aesthetics.  This should be faster and a tad more reliable.

Also fixed up some comments while here.

### DIFF
--- a/storage/objects.go
+++ b/storage/objects.go
@@ -55,7 +55,7 @@ func (s *ObjectsClient) Get(ctx context.Context, input *GetObjectInput) (*GetObj
 	}
 	respBody, respHeaders, err := s.client.ExecuteRequestStorage(ctx, reqInput)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


storage/objects.go
```diff
@@ -94,7 +94,7 @@ type DeleteObjectInput struct {
 
 // DeleteObject deletes an object.
 func (s *ObjectsClient) Delete(ctx context.Context, input *DeleteObjectInput) error {
-	path := fmt.Sprintf("/%s%s", s.client.AccountName, input.ObjectPath)
+	fullPath := path.Clean(path.Join("/", s.client.AccountName, input.ObjectPath))
```
cheapRoc: TIL `path`

> This package does not deal with Windows paths with drive letters or backslashes; to manipulate operating system paths, use the path/filepath package.

Which is exactly what we need here.

PR #69, opened 2017-12-11
Fix flaky list op

sean-:
`json.Decode()` can attempt to decode a short read. This would lead to strange list errors, like:

```
unable to list manta directory: Error decoding List response: invalid character ':' looking for beginning of value
```

Fortunately since the directory list operation returns newline-delimited JSON fragments, we can use a line-oriented scanner and unmarshal the desired entry reliably.

This PR assumes #68 has been applied.

### DIFF
--- a/storage/directory.go
+++ b/storage/directory.go
@@ -1,9 +1,9 @@
 package storage
 
 import (
+	"bufio"
 	"context"
 	"encoding/json"
-	"io"
 	"net/http"
 	"net/url"
 	"path"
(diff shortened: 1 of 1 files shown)


storage/directory.go
```diff
@@ -53,30 +53,30 @@ func (s *DirectoryClient) List(ctx context.Context, input *ListDirectoryInput) (
...
 	if err != nil {
-		return nil, errwrap.Wrapf("Error executing ListDirectory request: {{err}}", err)
+		return nil, errwrap.Wrapf("Error executing List request: {{err}}", err)
 	}
+	defer respBody.Close()
 
 	var results []*DirectoryEntry
-	for {
+	scanner := bufio.NewScanner(respBody)
+	for scanner.Scan() {
```
cheapRoc: Makes perfect sense since the split function scans lines by default.

sean-: My only issue with what I did there was there's no timeout or deadline cancellation. Something like this needs to be integrated still: https://gist.github.com/dchapes/6c992bf3e943934462509338cd213e99

PR #73, opened 2017-12-13
Move readers from an `io.ReadSeaker` to an `io.Reader`.

sean-:
When `triton-go` used a retriable HTTP client, it was required to use an `io.ReadSeaker`.  Now that `triton-go` doesn't use retriablehttp, relax this requirement to be just an `io.Reader`.  In the future we may revisit this and re-introduce a retriable client and promote this to an `io.ReadSeaker`, or figure out another way of passing in a Seaker.  The reason for wanting to do this is now you can read from a pipe (i.e. stdin).

This change is forward compatible.  Going back to an `io.ReadSeaker` will be backwards incompatible for people using non-seakable readers and will require additional thought.

Changelog entry: required

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -174,7 +174,7 @@ func (c *Client) ExecuteRequestURIParams(ctx context.Context, inputs RequestInpu
 	body := inputs.Body
 	query := inputs.Query
 
-	var requestBody io.ReadSeeker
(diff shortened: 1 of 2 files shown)


cheapRoc:
Ah right, we don't have any need to rewind the request body anymore. Good catch!

PR #75, opened 2017-12-14
storage: Introduce Objects.GetInfo for sending HEAD requests at Manta

### LINKED ISSUE #74: Support HEAD requests into Manta endpoints
This is a stub for adding HEAD request support for our Manta `storage` API. In Minio, we can't simply GET every file since it pulls the entire file body (which could be large). We need the ability to check information about objects in Manta.

Currently, this is supported by `node-manta` under the [`minfo`](https://github.com/joyent/node-manta/blob/master/bin/minfo) command. We'll want to replicate that here in our API as well.

/cc @stack72

cheapRoc:
This introduces two functions under the storage package, Objects.GetInfo and a convenience wrapper called Objects.IsDir. Objects.GetInfo performs the same action as Objects.Get except that it uses the http.MethodHead type request. All headers are returned akin to GetObjectOutput except for the response body. Objects.IsDir returns either a boolean, whether or not the object is a directory, and an error in case the Objects.GetInfo call failed.

Fixes #74

### DIFF
--- a/examples/storage/getobject.go
+++ b/examples/storage/getobject.go
@@ -14,6 +14,9 @@ import (
 	"github.com/joyent/triton-go/storage"
 )
 
+// This file stored in Manta is used in the example below.
+const path = "/stor/books/dracula.txt"
+
 func main() {
 	var (
 		signer authentication.Signer
@@ -27,26 +30,26 @@ func main() {
 	if keyMaterial == "" {
 		signer, err = authentication.NewSSHAgentSigner(keyID, accountName)
 		if err != nil {
-			log.Fatalf("Error Creating SSH Agent Signer: %s", err.Error())
+			log.Fatalf("error creating SSH agent signer: %v", err.Error())
 		}
 	} else {
 		var keyBytes []byte
 		if _, err = os.Stat(keyMaterial); err == nil {
 			keyBytes, err = ioutil.ReadFile(keyMaterial)
 			if err != nil {
-				log.Fatalf("Error reading key material from %s: %s",
+				log.Fatalf("error reading key material from %q: %v",
 					keyMaterial, err)
 			}
 			block, _ := pem.Decode(keyBytes)
 			if block == nil {
 				log.Fatalf(
-					"Failed to read key material '%s': no key found", keyMaterial)
+					"failed to read key material %q: no key found", keyMaterial)
 			}
 
 			if block.Headers["Proc-Type"] == "4,ENCRYPTED" {
-				log.Fatalf(
-					"Failed to read key '%s': password protected keys are\n"+
-						"not currently supported. Please decrypt the key prior to use.", keyMaterial)
+				log.Fatalf("failed to read key %q: password protected keys are\n"+
+					"not currently supported, decrypt key prior to use",
+					keyMaterial)
 			}
 
 		} else {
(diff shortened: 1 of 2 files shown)


storage/objects.go
```diff
@@ -19,6 +19,86 @@ type ObjectsClient struct {
...
+	}
+	response.Metadata = metadata
+
+	return response, nil
+}
+
+// IsDir is a convenience wrapper around the GetInfo function which takes an
+// ObjectPath and returns a boolean whether or not the object is a directory
+// type in Manta. Returns an error if GetInfo failed upstream for some reason.
+func (s *ObjectsClient) IsDir(ctx context.Context, objectPath string) (bool, error) {
```
cheapRoc: Normally I wouldn't like a helper that returns a `bool` to also include an error but the underlying HTTP request can fail and I want to detect that.

storage/objects.go
```diff
@@ -19,6 +19,86 @@ type ObjectsClient struct {
...
+}
+
+// GetInfo sends a HEAD request to an object in the Manta service. This function
+// does not return a response body.
+func (s *ObjectsClient) GetInfo(ctx context.Context, input *GetInfoInput) (*GetInfoOutput, error) {
+	fullPath := path.Clean(path.Join("/", s.client.AccountName, input.ObjectPath))
+
+	headers := &http.Header{}
+	for key, value := range input.Headers {
+		headers.Set(key, value)
```
sean-: Use `Add` instead of `Set` in order to allow multiple headers with the same name.  We should probably make a sweep for that exact usage elsewhere in the repo.

examples/storage/getobject.go
```diff
@@ -68,19 +70,54 @@ func main() {
...
+	isDir, err := client.Objects().IsDir(ctx, path)
+	if err != nil {
+		log.Fatalf("failed to detect directory '%s'\n", err)
+		return
+	}
+
+	if isDir {
+		fmt.Printf("'%s' is a directory\n", path)
+	} else {
+		fmt.Printf("'%s' is a file\n", path)
```
sean-: Can we convert both of these `%s` format verbs to `%q`?  Basically: `"'%s'` -> `"%q`  Faux quoting is a pet peeve, especially when we have `%q` floating around. :)

examples/storage/getobject.go
```diff
@@ -68,19 +70,54 @@ func main() {
...
+	}
+
+	ctx = context.Background()
+	obj, err := client.Objects().Get(ctx, &storage.GetObjectInput{
+		ObjectPath: path,
 	})
 	if err != nil {
-		log.Fatalf("compute.Objects.Get: %s", err)
+		log.Fatalf("failed to get '%s': %s", path, err)
```
sean-: `%v` for errors and use `%q` for the path.

storage/objects.go
```diff
@@ -19,6 +19,86 @@ type ObjectsClient struct {
...
+	}
+
+	reqInput := client.RequestInput{
+		Method:  http.MethodHead,
+		Path:    fullPath,
+		Headers: headers,
+	}
+	_, respHeaders, err := s.client.ExecuteRequestStorage(ctx, reqInput)
+	if err != nil {
+		return nil, errwrap.Wrapf("Error executing get info request: {{err}}", err)
```
sean-: This is a general style violation that we need to revisit throughout the library: error messages should begin with a lowercase letter, not uppercase.

examples/storage/getobject.go
cheapRoc: 👍 and you may have called me out on that before.

PR #77, opened 2017-12-14
Fix an issue where Manta HEAD requests do not return an error resp body

cheapRoc:


stack72:
Nice addition @cheapRoc 

We already saw this in action!

PR #78, opened 2017-12-15
Add support for recursively creating child directories

### LINKED ISSUE #61: Add support for Putting Objects of foo/bar
Rather than getting issues where `ParentNotDirectory` errors, we should add the ability to allow the user to specify a path as follows:

```
/account_name/stor/test-bucket/foo/bar
```

where foo is a directory that doesn't exist. The library should handle that logic. It looks similar as to the following:

```
objectPath := path.Join(mantaDefaultRootStore, m.directory)
	parts := strings.Split(entry.Key, "/")

	//This will remove the last part only, this way, I don't have to worry
	//about checking to see if the part is actually the last part
	parentParts := parts[:len(parts)-1] //foo
	keyPart := parts[len(parts)-1]      //bar

	for _, part := range parentParts {
		objectPath = path.Join(objectPath, part) //stor/test-bucket/foo
		err := m.client.Dir().Put(context.Background(), &storage.PutDirectoryInput{
			DirectoryName: objectPath,
		})
		if err != nil {
			return err
		}
	}
	return m.client.Objects().Put(context.Background(), &storage.PutObjectInput{
		ObjectPath:    path.Join(objectPath, keyPart),
		ObjectReader:  r,
		ContentLength: uint64(len(entry.Value)),
	})
```

stack72:
Fixes #61

when we want to Put an object in /stor/foo/bar and foo/bar doesn't exist, we will now create that object structure

```
% ./forcePut
2017/12/15 17:04:44 Folder Path "/stor/folder1/folder2/folder3/" not found
2017/12/15 17:04:44 Forcing the Creation of the Directory Tree: "/stor/folder1/folder2/folder3/foo.txt"
Successfully uploaded /tmp/foo.txt to /stor/folder1/folder2/folder3/foo.txt
```

### DIFF
--- a/examples/storage/forcePut.go
+++ b/examples/storage/forcePut.go
@@ -1,0 +1,88 @@
+package main
+
+import (
+	"encoding/pem"
+	"io/ioutil"
+	"log"
+	"os"
+
+	"context"
+
+	"fmt"
+
+	triton "github.com/joyent/triton-go"
+	"github.com/joyent/triton-go/authentication"
+	"github.com/joyent/triton-go/storage"
+)
+
+func main() {
+	keyID := os.Getenv("SDC_KEY_ID")
+	keyMaterial := os.Getenv("SDC_KEY_MATERIAL")
+	mantaUser := os.Getenv("MANTA_USER")
+
+	var signer authentication.Signer
+	var err error
+
+	if keyMaterial == "" {
+		signer, err = authentication.NewSSHAgentSigner(keyID, mantaUser)
+		if err != nil {
+			log.Fatalf("Error Creating SSH Agent Signer: %s", err.Error())
+		}
+	} else {
+		var keyBytes []byte
+		if _, err = os.Stat(keyMaterial); err == nil {
+			keyBytes, err = ioutil.ReadFile(keyMaterial)
+			if err != nil {
+				log.Fatalf("Error reading key material from %s: %s",
+					keyMaterial, err)
+			}
+			block, _ := pem.Decode(keyBytes)
+			if block == nil {
+				log.Fatalf(
+					"Failed to read key material '%s': no key found", keyMaterial)
+			}
+
+			if block.Headers["Proc-Type"] == "4,ENCRYPTED" {
+				log.Fatalf(
+					"Failed to read key '%s': password protected keys are\n"+
+						"not currently supported. Please decrypt the key prior to use.", keyMaterial)
+			}
+
+		} else {
+			keyBytes = []byte(keyMaterial)
+		}
+
+		signer, err = authentication.NewPrivateKeySigner(keyID, []byte(keyMaterial), mantaUser)
+		if err != nil {
+			log.Fatalf("Error Creating SSH Private Key Signer: %s", err.Error())
+		}
+	}
+
+	config := &triton.ClientConfig{
+		MantaURL:    os.Getenv("MANTA_URL"),
+		AccountName: mantaUser,
+		Signers:     []authentication.Signer{signer},
+	}
+
+	client, err := storage.NewClient(config)
+	if err != nil {
+		log.Fatalf("NewClient: %s", err)
+	}
+
+	reader, err := os.Open("/tmp/foo.txt")
+	if err != nil {
+		log.Fatalf("os.Open: %s", err)
+	}
+	defer reader.Close()
+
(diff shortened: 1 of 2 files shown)


examples/storage/forcePut.go
```diff
@@ -0,0 +1,87 @@
...
+		log.Fatalf("NewClient: %s", err)
+	}
+
+	reader, err := os.Open("/tmp/foo.txt")
+	if err != nil {
+		log.Fatalf("os.Open: %s", err)
+	}
+	defer reader.Close()
+
+	err = client.Objects().Put(context.Background(), &storage.PutObjectInput{
```
cheapRoc: Do we want a `ForceInsert` or something? Without it being optional we would wipe over whatever advantages we may have by reporting sub-directories as not existing in today's usage of Manta.

Thoughts?

storage/objects.go
```diff
@@ -258,6 +260,21 @@ type PutObjectInput struct {
 func (s *ObjectsClient) Put(ctx context.Context, input *PutObjectInput) error {
 	fullPath := path.Clean(path.Join("/", s.client.AccountName, input.ObjectPath))
 
+	err := checkDirectoryTreeExists(*s, ctx, input.ObjectPath)
+	if err != nil {
+		if strings.Contains(err.Error(), "ResourceNotFound") { //ResourceNotFound when dir.List doesn't exist
```
cheapRoc: I don't normally like this given a type system. We discussed error handling yesterday though so I'm fine with this for now.

storage/objects.go
```diff
@@ -304,3 +321,48 @@ func (s *ObjectsClient) Put(ctx context.Context, input *PutObjectInput) error {
...
+}
+
+func checkDirectoryTreeExists(c ObjectsClient, ctx context.Context, fullPath string) error {
+	dirClient := &DirectoryClient{
+		client: c.client,
+	}
+
+	folders, _ := path.Split(fullPath)
+
+	_, err := dirClient.List(ctx, &ListDirectoryInput{
```
cheapRoc: Any chance of using our new HEAD functions, `Objects.GetInfo` or `Objects.IsDir` here?

storage/objects.go
```diff
@@ -258,6 +260,21 @@ type PutObjectInput struct {
 func (s *ObjectsClient) Put(ctx context.Context, input *PutObjectInput) error {
 	fullPath := path.Clean(path.Join("/", s.client.AccountName, input.ObjectPath))
 
+	err := checkDirectoryTreeExists(*s, ctx, input.ObjectPath)
```
cheapRoc: If we're not checking within the loop how are we handling existing directories?

storage/objects.go
```diff
@@ -304,3 +321,48 @@ func (s *ObjectsClient) Put(ctx context.Context, input *PutObjectInput) error {
...
+		if i == 0 {
+			folderPath = part
+			continue
+		}
+		folderPath = path.Clean(path.Join(folderPath, part))
+		err := dirClient.Put(ctx, &PutDirectoryInput{
+			DirectoryName: folderPath,
+		})
+		if err != nil {
+			return err
```
cheapRoc: Will this error out if we can't create a directory along the way for whatever reason?

examples/storage/forcePut.go
stack72: Love this thought - much safer!

storage/objects.go
stack72: Yeah error handling needs to change all over the shop!

storage/objects.go
stack72: yes, it will error out and report that error to the user

storage/objects.go
stack72: This is a check to see if the folder path exists - not if it's a directory or an object

examples/storage/forcePut.go
stack72: Added :)

storage/objects.go
stack72: Demo'ed to @cheapRoc that we have a test case for this to work and works as expected

storage/objects.go
stack72: We have removed at least some of the string matching for now

storage/objects.go
stack72: Changed to using IsDir :)

PR #79, opened 2017-12-15
Clean up the handling of directory sanitization. Use abs paths everywhere

sean-:
The API allows for ambiguous input. Pass around absolute paths everywhere after crossing through the exported API.


stack72:
This now works as expected - thanks for the work here @sean-

cheapRoc:
👍

Issue #67, opened 2017-12-04
Manta client should support RBAC/Sub users


cheapRoc:
Ensure that RBAC ([some setup here](https://docs.joyent.com/public-cloud/rbac/quickstart#add-an-ssh-key-to-the-user)) works with `triton-go`'s Manta client support. Possibly support sub users by prepending a namespace of `/users` during key signing.

I'm told sub users work for Manta roles/policies and we should support that use case.

PR #82, opened 2017-12-28
Add support for rbac

stack72:
Fixes: #67 
Fixes: #83 

Adds support for RBAC via the ability to pass a Username to the SSH Agent signers

The examples and the associated README has been updated accordingly

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,11 +15,13 @@ using a key stored with the local SSH Agent (using an [`SSHAgentSigner`][6].
 To construct a Signer, use the `New*` range of methods in the `authentication`
 package. In the case of `authentication.NewSSHAgentSigner`, the parameters are
 the fingerprint of the key with which to sign, and the account name (normally
-stored in the `SDC_ACCOUNT` environment variable). For example:
+stored in the `SDC_ACCOUNT` environment variable). There is also support for 
+passing in a username, this will allow you to use an account other than the main
+Triton account. For example:
 
 ```
 const fingerprint := "a4:c6:f3:75:80:27:e0:03:a9:98:79:ef:c5:0a:06:11"
-sshKeySigner, err := authentication.NewSSHAgentSigner(fingerprint, "AccountName")
+sshKeySigner, err := authentication.NewSSHAgentSigner(fingerprint, "AccountName", "UserName")
 if err != nil {
     log.Fatalf("NewSSHAgentSigner: %s", err)
 }
@@ -40,6 +42,7 @@ the global `triton.ClientConfig` struct into the client's constructor function.
         TritonURL:   os.Getenv("SDC_URL"),
         MantaURL:    os.Getenv("MANTA_URL"),
         AccountName: accountName,
+        Username:    os.Getenv("SDC_USER"),
         Signers:     []authentication.Signer{sshKeySigner},
     }
 
@@ -87,7 +90,8 @@ set:
 
 Additionally, you may set `SDC_KEY_MATERIAL` to the contents of an unencrypted
 private key. If this is set, the PrivateKeySigner (see above) will be used - if
-not the SSHAgentSigner will be used.
+not the SSHAgentSigner will be used. You can also set `SDC_USER` to run the tests
+against an account other than the main Triton account
 
 ### Example Run
 
@@ -147,12 +151,13 @@ func main() {
     keyID := os.Getenv("SDC_KEY_ID")
     accountName := os.Getenv("SDC_ACCOUNT")
     keyMaterial := os.Getenv("SDC_KEY_MATERIAL")
+    userName := os.Getenv("SDC_USER")
 
     var signer authentication.Signer
     var err error
 
     if keyMaterial == "" {
-        signer, err = authentication.NewSSHAgentSigner(keyID, accountName)
+        signer, err = authentication.NewSSHAgentSigner(keyID, accountName, userName)
         if err != nil {
             log.Fatalf("Error Creating SSH Agent Signer: {{err}}", err)
         }
@@ -180,7 +185,7 @@ func main() {
             keyBytes = []byte(keyMaterial)
         }
 
-        signer, err = authentication.NewPrivateKeySigner(keyID, []byte(keyMaterial), accountName)
+        signer, err = authentication.NewPrivateKeySigner(keyID, []byte(keyMaterial), accountName, userName)
         if err != nil {
             log.Fatalf("Error Creating SSH Private Key Signer: {{err}}", err)
         }
@@ -189,6 +194,7 @@ func main() {
     config := &triton.ClientConfig{
         TritonURL:   os.Getenv("SDC_URL"),
         AccountName: accountName,
+        Username:    userName,
         Signers:     []authentication.Signer{signer},
     }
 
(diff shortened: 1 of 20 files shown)


README.md
```diff
@@ -15,11 +15,13 @@ using a key stored with the local SSH Agent (using an [`SSHAgentSigner`][6].
 To construct a Signer, use the `New*` range of methods in the `authentication`
 package. In the case of `authentication.NewSSHAgentSigner`, the parameters are
 the fingerprint of the key with which to sign, and the account name (normally
-stored in the `SDC_ACCOUNT` environment variable). For example:
+stored in the `SDC_ACCOUNT` environment variable). There is also support for
```
sean-: Alright, when do we want to have our watershed moment and move from `SDC_ACCOUNT` in docs to `TRITON_ACCOUNT` in docs?  This isn't an issue with this PR, however.  Opening joyent/triton-go#83 to track this issue.

authentication/private_key_signer.go
```diff
@@ -20,12 +20,13 @@ type PrivateKeySigner struct {
...
 	algorithm               string
 	accountName             string
+	userName                string
 	hashFunc                crypto.Hash
 
 	privateKey *rsa.PrivateKey
 }
 
-func NewPrivateKeySigner(keyFingerprint string, privateKeyMaterial []byte, accountName string) (*PrivateKeySigner, error) {
+func NewPrivateKeySigner(keyFingerprint string, privateKeyMaterial []byte, accountName, userName string) (*PrivateKeySigner, error) {
```
sean-: Sweet, this is a good breaking change.  Changelog++  I'm assuming you, @stack72 , are going to chase the common breakage in tooling? https://godoc.org/github.com/joyent/triton-go?importers

btw, the reason I'm okay with this API change is because the API should specify the username that has signed the request.  Implicit identity is "problematic."

I'd rather suck up the explicit breakage now vs introduce something using an ellipsis that we later have to rip out.

authentication/private_key_signer.go
```diff
@@ -80,7 +85,12 @@ func (s *PrivateKeySigner) Sign(dateHeader string) (string, error) {
 	}
 	signedBase64 := base64.StdEncoding.EncodeToString(signed)
 
-	keyID := fmt.Sprintf("/%s/keys/%s", s.accountName, s.formattedKeyFingerprint)
+	var keyID string
+	if s.userName != "" {
+		keyID = fmt.Sprintf("/%s/users/%s/keys/%s", s.accountName, s.userName, s.formattedKeyFingerprint)
```
sean-: Let's change these to calls to `path.Join()`.

authentication/ssh_agent_signer.go
```diff
@@ -55,7 +56,12 @@ func NewSSHAgentSigner(keyFingerprint, accountName string) (*SSHAgentSigner, err
 	}
 	signer.key = matchingKey
 	signer.formattedKeyFingerprint = formatPublicKeyFingerprint(signer.key, true)
-	signer.keyIdentifier = fmt.Sprintf("/%s/keys/%s", signer.accountName, signer.formattedKeyFingerprint)
+	if userName != "" {
+		signer.userName = userName
+		signer.keyIdentifier = fmt.Sprintf("/%s/users/%s/keys/%s", signer.accountName, userName, signer.formattedKeyFingerprint)
+	} else {
+		signer.keyIdentifier = fmt.Sprintf("/%s/keys/%s", signer.accountName, signer.formattedKeyFingerprint)
```
sean-: Same here, `path.Join()`.

authentication/private_key_signer.go
sean-: ... you know, since we're making a breaking change, do we want to just outright break it and use an input struct?

```go
type NewPrivateKeySignerInput struct {
  KeyFingerPrint string
  PrivateKeyMaterial []byte
  AccountName string
  Username string
}
```

And pass that in (by value, ideally)?

This would make the API future-proof from additions like delegation. ?

examples/compute/create_instance.go
```diff
@@ -147,4 +149,12 @@ func main() {
...
 		fmt.Println("Timed out")
 	}
+
+	fmt.Println("Cleaning up machine....")
+	err = c.Instances().Delete(context.Background(), &compute.DeleteInstanceInput{
+		ID: created.ID,
+	})
+	if err != nil {
+		log.Fatalf("Delete(): %v", err)
+	}
```
sean-: Nice addition to the examples.

testutils/runner.go
```diff
@@ -38,6 +38,7 @@ func AccTest(t *testing.T, c TestCase) {
 	sdcAccount := os.Getenv("SDC_ACCOUNT")
 	sdcKeyId := os.Getenv("SDC_KEY_ID")
 	sdcKeyMaterial := os.Getenv("SDC_KEY_MATERIAL")
+	sdcUser := os.Getenv("SDC_USER")
```
sean-: `sdcUser` -> `userName` for consistency.

```sh
git grep -i sdc . > /dev/null
```

README.md
stack72: I will follow up with a PR to solve this

authentication/private_key_signer.go
stack72: I really like this idea! I will change to that

README.md
jen20: We may as well do this now too.

authentication/private_key_signer.go
jen20: :+1:, this matches the rest of the API at that point too.

README.md
sean-: I would advocate for that before we tag `0.4.0`, yes.

authentication/private_key_signer.go
```diff
@@ -20,15 +22,23 @@ type PrivateKeySigner struct {
...
+	userName                string
 	hashFunc                crypto.Hash
 
 	privateKey *rsa.PrivateKey
 }
 
-func NewPrivateKeySigner(keyFingerprint string, privateKeyMaterial []byte, accountName string) (*PrivateKeySigner, error) {
-	keyFingerprintMD5 := strings.Replace(keyFingerprint, ":", "", -1)
+type PrivateKeySignerInput struct {
+	KeyFingerPrint     string
```
sean-: I missed this a minute ago, @stack72 .  This member should be renamed to be `KeyFingerprint` or even simplified down to just `KeyID`.

README.md
```diff
@@ -15,11 +15,17 @@ using a key stored with the local SSH Agent (using an [`SSHAgentSigner`][6].
 To construct a Signer, use the `New*` range of methods in the `authentication`
 package. In the case of `authentication.NewSSHAgentSigner`, the parameters are
 the fingerprint of the key with which to sign, and the account name (normally
-stored in the `SDC_ACCOUNT` environment variable). For example:
+stored in the `TRITON_ACCOUNT` environment variable). There is also support for
```
sean-: Trailing space here on this line, too.

authentication/private_key_signer.go
```diff
@@ -11,6 +11,8 @@ import (
 	"fmt"
 	"strings"
 
+	"path"
```
sean-: Nuke the whitespace above this emport and re-run `goimport`.

authentication/ssh_agent_signer.go
```diff
@@ -10,6 +10,8 @@ import (
 	"os"
 	"strings"
 
+	"path"
```
sean-: Nuke whitespace above import

PR #85, opened 2017-12-28
Signer API nits...

sean-:
Dept of 2nd thoughts kicked in when reviewing the changes in a `dep ensure -update && dep prune`.  This PR is a stab at fixing most of the loose items.


stack72:
LGTM! I agree with all of this - thanks for this :)

Issue #76, opened 2017-12-14
Lowercase and format error messages properly


cheapRoc:
As noted in PR #75, the following style rules should be adjusted across the project.

- Error messages should begin with a lowercase letter, not uppercase.
- Use `%v` format verb when printing errors.
- Use `%q` for any quoted string.

I've made an attempt to introduce these under `examples/storage/getobject.go` but we'll want them project wide when someone has a moment.

PR #87, opened 2018-01-04
account pkg: To remove errwrap, add unit testing and enhance error messages

stack72:
Part of: #10 
Part of: #76 

- [x] Account Package
  - [x] Unit tests
  - [x] Error Messages


Gopkg.toml
```diff
@@ -40,3 +40,7 @@
 [[constraint]]
   branch = "master"
   name = "golang.org/x/crypto"
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/pkg/errors"
```
sean-: Missing newline

Gopkg.lock
```diff
@@ -7,17 +7,23 @@
   packages = ["."]
   revision = "d5467c17e7afe8d8f08f556c6c811a50c3feb28d"
 
+[[projects]]
+  name = "github.com/davecgh/go-spew"
+  packages = ["spew"]
+  revision = "346938d642f2ec3594ed81d874461961cd0faa76"
+  version = "v1.1.0"
```
sean-: Why is `spew` being used?  I'm guessing there's some uncommitted code locally that `dep` picked up.

account/config.go
```diff
@@ -55,7 +55,7 @@ type UpdateConfigInput struct {
 func (c *ConfigClient) Update(ctx context.Context, input *UpdateConfigInput) (*Config, error) {
 	path := fmt.Sprintf("/%s/config", c.client.AccountName)
 	reqInputs := client.RequestInput{
-		Method: http.MethodPut,
+		Method: http.MethodPost,
```
sean-: 1) Why the change from `PUT` to `POST`?

2) Someday we should come through and replace the `fmt.Sprintf()` calls with `path.Join("/", c.client.AccountName, "config"`.  Not now, but at some point...

Gopkg.lock
stack72: I believe this is actually part of Errwrap dependencies - when we remove errwrap, it will go away

stack72: Oh maybe not....

account/config.go
stack72: So the change from PUT to POST was to follow the convention of the other Update API endpoints in this package which are all POST

stack72: I reverted this - I am not sure this is the correct thing to do - I shouldn't change an endpoint that is already in place :)

PR #90, opened 2018-01-08
Remove errwrap from storage pkg

stack72:
Right now, I am not going to add tests for manta - they are proving to not be a pretty thing - I will come back and update these in the future

### DIFF
--- a/storage/snaplink_test.go
+++ b/storage/snaplink_test.go
@@ -1,0 +1,79 @@
+package storage_test
+
+import (
+	"context"
+	"net/http"
+	"path"
+	"strings"
+	"testing"
+
+	"github.com/joyent/triton-go/storage"
+	"github.com/joyent/triton-go/testutils"
+	"github.com/pkg/errors"
+)
+
+const accountUrl = "testing"
+
+var (
+	putSnapLinkErrorType = errors.New("unable to put snaplink")
+	linkPath             = "/stor/foobar.json"
+	brokenLinkPath       = "/missingfolder/foo.json"
+	sourcePath           = "/stor/foo.json"
+)
+
+func MockStorageClient() *storage.StorageClient {
+	return &storage.StorageClient{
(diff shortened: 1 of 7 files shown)


cheapRoc:
Great having more unit tests and switching out errors, couple comments.

What happens to the `client/errors.go` and `compute/errors.go` files? More specifically how are we going to migrate away from [this `errwrap` functionality](https://github.com/joyent/triton-go/blob/master/client/errors.go#L179) that identifies specific errors? I feel like that was something we should have had a shim for before we started merging into master.

I'd like to split out the acceptance tests from the unit tests. I don't mind them being mixed until we do that though. I started that on a branch but that's old now. Not a big deal.

Also, looks like a test binary got committed at `examples/storage/list_directory/main`. But sounds like maybe this PR isn't finished anyway?

storage/snaplink_test.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+	})
+}
+
+func putSnapLinkSuccess(req *http.Request) (*http.Response, error) {
+	header := http.Header{}
+	header.Add("Content-Type", "application/json")
+
+	return &http.Response{
+		StatusCode: 204,
```
sean-: `204` -> `http.StatusNoContent`

storage/snaplink_test.go
```diff
@@ -0,0 +1,79 @@
...
+	t.Run("error", func(t *testing.T) {
+		testutils.RegisterResponder("PUT", path.Join("/", accountUrl, brokenLinkPath), putSnapLinkError)
+
+		err := do(context.Background(), storageClient)
+		if err == nil {
+			t.Fatal(err)
+		}
+
+		if !strings.Contains(err.Error(), "unable to put snaplink") {
+			t.Errorf("expected error to equal testError: found %s", err)
```
sean-: `%v` for the fmt verb.

PR #94, opened 2018-01-11
pkg/compute: Replacing errwrap in the compute package

stack72:


stack72:
merged this PR into #95

Issue #92, opened 2018-01-09
New errors package


cheapRoc:
There exists a fair bit of overlap in terms of error handling which came out of merging the two libraries that made up the SDK, `triton-go` and `manta-go`. There are two error types, `TritonError` and `MantaError` as well as two files that came out of the merge `client/errors.go` and `compute/errors.go`. Their placement within the SDK was largely due to where the errors were imported.

@stack72 is planning to create a `triton-go/errors` package which will merge the two types and files as much as possible. This issue is tracking that package specifically.

PR #95, opened 2018-01-11
pkg/errors: Exposing all of the Errors in 1 specific package

stack72:
Fixes: #92
Fixes: #97 
Fixes: #76 
Fixes: #10

This consolidates the work of the errors in client/errors.go and compute/errors.go

We have a specific type of errors we can check for now

please note the alias of the github.com/joyent/triton-go/errors package to `tt`. This is because we already use the github.com/pkg/errors package


client/client.go
```diff
@@ -169,7 +169,7 @@ func doNotFollowRedirects(*http.Request, []*http.Request) error {
 // 	return fmt.Sprintf("%s%s", c.Endpoint, path)
 // }
 
-func (c *Client) DecodeError(statusCode int, body io.Reader) error {
+func (c *Client) DecodeEror(statusCode int, body io.Reader) error {
```
cheapRoc: This misspelling wasn't intentional, correct?

errors/errors.go
```diff
@@ -173,17 +160,76 @@ func IsUserDoesNotExistError(err error) bool {
...
-// an underlying MantaError with code errorCode.
-func isSpecificError(err error, errorCode string) bool {
-	tritonErrorInterface := errwrap.GetType(err.(error), &MantaError{})
-	if tritonErrorInterface == nil {
-		return false
-	}
+func IsBadRequest(err error) bool {
+	return isSpecificError(err, "BadRequest")
+}
+
```
cheapRoc: Can you make sure the error funcs in this file are either one space apart or have no spaces between them, not both? Just for the sake of consistency.

client/client.go
```diff
@@ -16,6 +16,7 @@ import (
 
 	"github.com/hashicorp/errwrap"
 	"github.com/joyent/triton-go/authentication"
+	tt "github.com/joyent/triton-go/errors"
```
sean-: I don't like `tt` as a package alias for `errors`.  I'd do something more verbose like `tritonErrors` if possible.  I don't like "`terrors`", either, because that becomes confusing to the reader.

Actually, is this awful?

```go
import (
  stderrors "errors"

  "github.com/joyent/triton-go/errors"
)
```

Things like `fmt.Errorf()` are unaffected and it makes it more difficult to introduce a non-standard error.

We could even change `New()` so that calls like:

```
&tt.TritonError{
  StatusCode: response.StatusCode,
￼  Code:       "ResourceNotFound",
}
```

can be replaced by:

```
errors.New(response.StatusCode, "ResourceNotFound")
```

Would you be okay with that?  I'm still mulling the stutter of `tritonError.TritonError` and `StatusCode` + `Code`.  Can those two be folded into the constructor so that `triton-go/errors/New()` derives the `Code` automatically based on the `response.StatusCode`?  Maybe `NewFromStatusCode(response.StatusCode)` does the lookup?

compute/images.go
```diff
@@ -10,6 +10,7 @@ import (
 
 	"github.com/hashicorp/errwrap"
 	"github.com/joyent/triton-go/client"
+	tt "github.com/joyent/triton-go/errors"
```
sean-: -> `tritonErrors`

errors/errors.go
```diff
@@ -0,0 +1,276 @@
+package errors
```
sean-: A copyright header block is needed here.

errors/errors.go
```diff
@@ -0,0 +1,276 @@
...
+	"fmt"
+
+	"github.com/pkg/errors"
+)
+
+// TritonError represents an error code and message along with
+// the status code of the HTTP request which resulted in the error
+// message. Error codes used by the Triton API are listed at
+// https://apidocs.joyent.com/cloudapi/#cloudapi-http-responses
+type TritonError struct {
```
sean-: Can we do a `go-rename` on `TritonError` and change it to be just `Triton` ?  The struct type is already scoped to the package name.  Future type assertions would read: `errors.Triton` or however the consumer aliases `triton-go/errors`.  ?

errors/errors.go
```diff
@@ -0,0 +1,276 @@
...
+// Error implements interface Error on the MantaError type.
+func (e MantaError) Error() string {
+	return fmt.Sprintf("%s: %s", e.Code, e.Message)
+}
+
+// MantaError represents an error code and message along with
+// the status code of the HTTP request which resulted in the error
+// message. Error codes used by the Manta API are listed at
+// https://apidocs.joyent.com/manta/api.html#errors
+type MantaError struct {
```
sean-: Same comment but for `MantaError`.  ?

.. actually, evolving thought process here.  Is this thrust correct?  I understand wanting to centralize the client errors into a single package, that makes sense.  But in terms of responses that come back from the API, I'm not sure this is right.  A Compute error may share a message with a Manta error, but the two aren't interchangeable types.  Wouldn't we want to do something like:

- Add `compute/errors.go`
- Add `compute.Error` as a type
- Add `manta.Error` as a type

Different packages are going to have differing levels of detail they may want to expose, but in all cases, none of the error types are fungable.  We may want to have it so that the client exposes an error type that satisfies a `TritonError` `interface` that can be type-asserted into `compute.Error` or `manta.Error`, but this feels like we're trying to create a single straight-jacket for all errors to fit into.

And to be clear, the current error handling has been sufficient, but isn't ideal, so I'm glad we're doing this.  But I'm not sure we're doing the right thing quite yet.

I just went to look at what the Google API client does (they seem like an informed bunch w.r.t. error interface design... hrm...):

https://github.com/google/google-api-go-client/blob/a11cf47576f3bc79d7e517bf21887f4c390b1a69/googleapi/googleapi.go#L61

Yeah, that seems passable.  The have a slice of errors from the API.  All APIs return a consistent error type.  Each service endpoint, however, doesn't have its own error type:

https://github.com/google/google-api-go-client/blob/3a1d936b7575b82197a1fea0632218dd07b1e65c/pubsub/v1/pubsub-gen.go#L1477-L1495

I wonder if that's because this code base is transpiled from some other service definition.  Let's take this offline.  I'm not entirely convinced of my own arguments and am open to debate.

My instinct says each discrete component of Triton should have its own error type with extra details and that each component error type should implement a standard interface so that clients don't need to be bothered with the details unless they want the supplemental details that `compute` may provide, for example.

storage/objects.go
```diff
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/joyent/triton-go/client"
+	tt "github.com/joyent/triton-go/errors"
```
sean-: Following up on previous comments:

a `storage.Error` could embed `triton-go/errors.CommonError` (or some standard package that implements the standard `TritonError` `interface`).

sean-:
@stack72 talked offline and we are going to implement a top-level `APIError` that will satisfy the `errors` interface.  There will be two main "error" types going forward, `APIError` (received something back from an API endpoint) and `ClientError` (local or network related errors). Individual packages will have the opportunity to satisfy the `APIError` interface and provide additional detail as necessary.  My only additional request is that the `APIError` interface include a `Type() string` method so that it's easy to figure out from an `APIError` what the underlying type actually is.  Consumers of `triton-go` will only be presented with `errors`.  Additional details will show up in the next PR.

+1 for doing this work @stack72.

stack72:
We now get API errors (for calls that interact with Triton or Manta) in the following format

```
0.000s err=&errors.withStack{
           error: &errors.withMessage{
               cause: &errors.APIError{StatusCode:404, Code:"ResourceNotFound", Message:"/stack72_joyent/stor/missing-folder was not found"},
               msg:   "unable to list directory",
           },
           stack: &errors.stack{0x126132b, 0x129da21, 0x102cab6, 0x10580a1},
       }
```

This means we can now test for specific error types in our code:

```
ctx := context.Background()
    output, err := client.Dir().List(ctx, &storage.ListDirectoryInput{
        DirectoryName: path,
    })
    if err != nil {
        if errors.IsResourceNotFound(err) {
            fmt.Print("Correct Error achieved\n")
            return
        }
        return
    }
```

This gives us the following output:

```
list_directory [errors-lkg●] % go run main.go
Correct Error achieved
```

We can even flatten the error and get the following output:

```
list_directory [errors-lkg●] % go run main.go
Correct Error achieved
unable to list directory: ResourceNotFound: /stack72_joyent/stor/missing-folder was not found
```

Nothing has been done with ClientError in the interim

compute/ping.go
```diff
@@ -3,11 +3,10 @@ package compute
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"net/http"
 
-	"github.com/hashicorp/errwrap"
 	"github.com/joyent/triton-go/client"
+	stderrors "github.com/pkg/errors"
```
sean-: Across all files here, `s/stderrors/pkgerrors/g`

compute/ping_test.go
```diff
@@ -74,37 +72,11 @@ func TestPing(t *testing.T) {
 			t.Error("expected pingOut to be nil")
 		}
 
-		if err.Error() != blankError {
+		if !strings.Contains(err.Error(), "unable to ping") {
```
sean-: Is this right?  Is there a different helper method we should be using instead?  This seems like a grey area, so I have no idea.

compute/services_test.go
```diff
@@ -1,12 +1,19 @@
...
+package compute_test
 
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
+	"net/http"
 	"testing"
 
+	"path"
```
sean-: whitespace

errors/errors.go
```diff
@@ -0,0 +1,272 @@
...
+// https://apidocs.joyent.com/manta/api.html#errors
+type APIError struct {
+	StatusCode int
+	Code       string `json:"code"`
+	Message    string `json:"message"`
+}
+
+// Error implements interface Error on the APIError type.
+func (e APIError) Error() string {
+	return fmt.Sprintf("%s: %s", e.Code, e.Message)
```
sean-: Safe guard against garbage in log messages or errors:

```go
return strings.Trim(fmt.Sprintf("%+q", e.Code), `"`) + ": " + strings.Trim(fmt.Sprintf("%+q", e.Message), `"`)
```

errors/errors.go
```diff
@@ -0,0 +1,272 @@
...
+// when connecting to the triton-go client
+type ClientError struct {
+	StatusCode int
+	Code       string `json:"code"`
+	Message    string `json:"message"`
+}
+
+// Error implements interface Error on the ClientError type.
+func (e ClientError) Error() string {
+	return fmt.Sprintf("%s: %s", e.Code, e.Message)
```
sean-: Same thing here.

version.go
```diff
@@ -0,0 +1,22 @@
+package triton
+
+import (
+	"fmt"
+	"runtime"
+)
+
+// The main version number of the current released Triton-go SDK.
+const Version = "1.0.0"
```
sean-: Can we jump to `"0.9.0"` for a week before calling it `1.0.0` so that we can integrate this into other tools and make sure we don't need to make any other tweaks?

version.go
```diff
@@ -0,0 +1,22 @@
...
+
+// A pre-release marker for the version. If this is "" (empty string)
+// then it means that it is a final release. Otherwise, this is a pre-release
+// such as "dev" (in development), "beta", "rc1", etc.
+var Prerelease = "dev"
```
sean-: Can you include `go build` linker instructions to denote a release?  I'd rather flip this around so that it's not `"dev"` by default because this is a library.  If something wants to add "dev" to the useragent because the consuming application is still under development, that makes sense to me.  But doing it here seems like it'll be a PITA to root through the code to change.  My worry being that everything will be decorated with `triton-go/1.0.0-dev` unless they dig through and provide the right linker incantation to swap out this variable (e.g. think about what'd be involved in changing this in the Terraform provider).

compute/images.go
```diff
@@ -163,10 +162,9 @@ func (c *ImagesClient) Export(ctx context.Context, input *ExportImageInput) (*Ma
 	fullPath := path.Join("/", c.client.AccountName, "images", input.ImageID)
 	query := &url.Values{}
 	query.Set("action", "export")
-	query.Set("manta_path", input.MantaPath)
 
 	reqInputs := client.RequestInput{
-		Method: http.MethodGet,
+		Method: http.MethodPost,
```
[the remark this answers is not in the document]
stack72: Yes, this was an error I found with the addition of tests :)

compute/ping_test.go
stack72: Not for this specific error type - this is something we will wrap in future i'd say

version.go
stack72: I don't know go build linker but i will remove the string for now :)

errors/errors.go
sean-: Punted to #98

compute/images.go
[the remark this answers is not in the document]
sean-: Nice!

errors/errors.go
```diff
@@ -3,6 +3,8 @@ package errors
 import (
 	"fmt"
 
+	"strings"
```
sean-: I'm giving up on GoLand's `goimport` functionality, it's broken and needs to use the `goimports` utility written in Go.  Something to add to your feature request along with official `gorename` support.  :~]

version.go
```diff
@@ -20,3 +20,5 @@ func UserAgent() string {
 		return fmt.Sprintf("triton-go/%s (%s-%s; %s)", Version, runtime.GOARCH, runtime.GOOS, runtime.Version())
 	}
 }
+
+const AcceptVersion = "8"
```
sean-: `s/AcceptVersion/CloudAPIMajorVersion/`

```
// The minimum CloudAPI version we will accept with this version of joyent/triton-go
const CloudAPIMajorVersionStr = "8"
```

sean-:
NICE!  Thank you @stack72 / @cheapRoc !!!

PR #96, opened 2018-01-11
pkg/client: Removal of errwrap from the client package

stack72:


stack72:
Combined with #95 as it makes more sense to do this

PR #100, opened 2018-01-12
compute/volumes: Add support for volume management in Triton

stack72:
This will allow the CRUD functionality of volumes when they are
enabled in each of the environments. Right now, it is a whitelist
feature only

### DIFF
--- a/compute/volumes.go
+++ b/compute/volumes.go
@@ -1,0 +1,216 @@
+//
+// Copyright (c) 2018, Joyent, Inc. All rights reserved.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+
+package compute
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/url"
+	"path"
+
+	"github.com/joyent/triton-go/client"
+	"github.com/pkg/errors"
+)
+
+type VolumesClient struct {
+	client *client.Client
+}
+
+type Volume struct {
+	ID             string   `json:"id"`
+	Name           string   `json:"name"`
+	Owner          string   `json:"owner_uuid"`
+	Type           string   `json:"type"`
+	FileSystemPath string   `json:"filesystem_path"`
+	State          string   `json:"state"`
+	Networks       []string `json:"networks"`
+	Refs           []string `json:"refs"`
+}
+
+type ListVolumesInput struct {
+	Name  string
+	Size  string
+	State string
+	Type  string
+}
+
+func (c *VolumesClient) List(ctx context.Context, input *ListVolumesInput) ([]*Volume, error) {
+	fullPath := path.Join("/", c.client.AccountName, "volumes")
+
+	query := &url.Values{}
+	if input.Name != "" {
+		query.Set("name", input.Name)
+	}
+	if input.Size != "" {
+		query.Set("size", input.Size)
+	}
+	if input.State != "" {
+		query.Set("state", input.State)
+	}
+	if input.Type != "" {
+		query.Set("type", input.Type)
+	}
+
+	reqInputs := client.RequestInput{
+		Method: http.MethodGet,
+		Path:   fullPath,
+		Query:  query,
+	}
+	resp, err := c.client.ExecuteRequest(ctx, reqInputs)
+	if resp != nil {
+		defer resp.Close()
+	}
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to list volumes")
+	}
+
+	var result []*Volume
+	decoder := json.NewDecoder(resp)
+	if err = decoder.Decode(&result); err != nil {
+		return nil, errors.Wrap(err, "unable to decode list volumes response")
+	}
+
+	return result, nil
+}
+
+type CreateVolumeInput struct {
+	Name     string
+	Size     int64
+	Networks []string
+	Type     string
+}
+
+func (input *CreateVolumeInput) toAPI() map[string]interface{} {
+	result := make(map[string]interface{}, 0)
+
+	if input.Name != "" {
+		result["name"] = input.Name
+	}
+
+	if input.Size != 0 {
+		result["size"] = input.Size
+	}
+
+	if input.Type != "" {
+		result["type"] = input.Type
+	}
+
+	if len(input.Networks) > 0 {
+		result["networks"] = input.Networks
+	}
+
+	return result
+}
+
+func (c *VolumesClient) Create(ctx context.Context, input *CreateVolumeInput) (*Volume, error) {
+	fullPath := path.Join("/", c.client.AccountName, "volumes")
+
+	reqInputs := client.RequestInput{
+		Method: http.MethodPost,
+		Path:   fullPath,
+		Body:   input.toAPI(),
+	}
+	resp, err := c.client.ExecuteRequest(ctx, reqInputs)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to create volume")
+	}
+	if resp != nil {
+		defer resp.Close()
+	}
+
+	var result *Volume
+	decoder := json.NewDecoder(resp)
+	if err = decoder.Decode(&result); err != nil {
+		return nil, errors.Wrap(err, "unable to decode create volume response")
+	}
+
+	return result, nil
+}
+
+type DeleteVolumeInput struct {
+	ID string
+}
+
+func (c *VolumesClient) Delete(ctx context.Context, input *DeleteVolumeInput) error {
+	fullPath := path.Join("/", c.client.AccountName, "volumes", input.ID)
+	reqInputs := client.RequestInput{
+		Method: http.MethodDelete,
+		Path:   fullPath,
+	}
+	resp, err := c.client.ExecuteRequestRaw(ctx, reqInputs)
+	if err != nil {
+		return errors.Wrap(err, "unable to delete volume")
+	}
+	if resp == nil {
+		return errors.Wrap(err, "unable to delete volume")
+	}
+	if resp.Body != nil {
+		defer resp.Body.Close()
+	}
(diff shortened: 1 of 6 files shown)


compute/volumes.go
```diff
@@ -0,0 +1,216 @@
...
+	reqInputs := client.RequestInput{
+		Method: http.MethodGet,
+		Path:   fullPath,
+		Query:  query,
+	}
+	respReader, err := c.client.ExecuteRequest(ctx, reqInputs)
+	if respReader != nil {
+		defer respReader.Close()
+	}
+	if err != nil {
```
sean-: Move `err != nil` above `respReader`

compute/volumes.go
```diff
@@ -0,0 +1,216 @@
...
+	reqInputs := client.RequestInput{
+		Method: http.MethodPost,
+		Path:   fullPath,
+		Body:   input.toAPI(),
+	}
+	respReader, err := c.client.ExecuteRequest(ctx, reqInputs)
+	if respReader != nil {
+		defer respReader.Close()
+	}
+	if err != nil {
```
sean-: Ditto moving this.

compute/volumes.go
```diff
@@ -0,0 +1,216 @@
...
+	if response == nil {
+		return errors.Wrap(err, "unable to delete volume")
+	}
+	if response.Body != nil {
+		defer response.Body.Close()
+	}
+	if response.StatusCode == http.StatusNotFound || response.StatusCode == http.StatusGone {
+		return nil
+	}
+	if err != nil {
```
sean-: Move before the `response.Body != nil` check.  Is there a case here where `resp` can be not `nil` error not `nil`?  I thought they were mutually exclusive here.  ?

compute/volumes.go
```diff
@@ -0,0 +1,216 @@
...
+	ID string
+}
+
+func (c *VolumesClient) Delete(ctx context.Context, input *DeleteVolumeInput) error {
+	fullPath := path.Join("/", c.client.AccountName, "volumes", input.ID)
+	reqInputs := client.RequestInput{
+		Method: http.MethodDelete,
+		Path:   fullPath,
+	}
+	response, err := c.client.ExecuteRequestRaw(ctx, reqInputs)
```
sean-: `s/response/resp/g` to be consistent with nearly all other users of `req`/`resp`. ?

examples/compute/volumes/crud/main.go
```diff
@@ -0,0 +1,154 @@
...
+package main
+
+import (
+	"context"
+	"encoding/pem"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+
```
sean-: GoLand: demerit.

examples/compute/volumes/crud/main.go
```diff
@@ -0,0 +1,154 @@
...
+	"time"
+
+	"github.com/joyent/triton-go"
+	"github.com/joyent/triton-go/authentication"
+	"github.com/joyent/triton-go/compute"
+)
+
+func main() {
+	keyID := os.Getenv("SDC_KEY_ID")
```
sean-: I know this is an example, but `SDC_*` ?  Why not `TRITON_*`?

examples/compute/volumes/list_volumes/main.go
```diff
@@ -0,0 +1,101 @@
...
+
+	if keyMaterial == "" {
+		input := authentication.SSHAgentSignerInput{
+			KeyID:       keyID,
+			AccountName: accountName,
+			Username:    userName,
+		}
+		signer, err = authentication.NewSSHAgentSigner(input)
+		if err != nil {
+			log.Fatalf("Error Creating SSH Agent Signer: %s", err.Error())
```
sean-: `%v`.  Do we need the `.Error()` method call to be explicitly called here?

examples/compute/volumes/list_volumes/main.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+	} else {
+		var keyBytes []byte
+		if _, err = os.Stat(keyMaterial); err == nil {
+			keyBytes, err = ioutil.ReadFile(keyMaterial)
+			if err != nil {
+				log.Fatalf("Error reading key material from %s: %s",
```
sean-: `%v`

examples/compute/volumes/list_volumes/main.go
```diff
@@ -0,0 +1,101 @@
...
+					keyMaterial, err)
+			}
+			block, _ := pem.Decode(keyBytes)
+			if block == nil {
+				log.Fatalf(
+					"Failed to read key material '%s': no key found", keyMaterial)
```
sean-: `s/'%s'/%q/`

examples/compute/volumes/list_volumes/main.go
```diff
@@ -0,0 +1,101 @@
...
+			}
+
+			if block.Headers["Proc-Type"] == "4,ENCRYPTED" {
+				log.Fatalf(
+					"Failed to read key '%s': password protected keys are\n"+
```
sean-: -> `%q`

examples/compute/volumes/list_volumes/main.go
```diff
@@ -0,0 +1,101 @@
...
+
+		input := authentication.PrivateKeySignerInput{
+			KeyID:              keyID,
+			PrivateKeyMaterial: keyBytes,
+			AccountName:        accountName,
+			Username:           userName,
+		}
+		signer, err = authentication.NewPrivateKeySigner(input)
+		if err != nil {
+			log.Fatalf("Error Creating SSH Private Key Signer: %s", err.Error())
```
sean-: `%v`

examples/compute/volumes/list_volumes/main.go
```diff
@@ -0,0 +1,101 @@
...
+	config := &triton.ClientConfig{
+		TritonURL:   os.Getenv("SDC_URL"),
+		AccountName: accountName,
+		Username:    userName,
+		Signers:     []authentication.Signer{signer},
+	}
+
+	c, err := compute.NewClient(config)
+	if err != nil {
+		log.Fatalf("compute.NewClient: %s", err)
```
sean-: `%v`

examples/compute/volumes/list_volumes/main.go
```diff
@@ -0,0 +1,101 @@
...
+
+	listInput := &compute.ListVolumesInput{}
+	volumes, err := c.Volumes().List(context.Background(), listInput)
+	if err != nil {
+		log.Fatalf("compute.Volumes.List: %v", err)
+	}
+	numInstances := 0
+	for _, volume := range volumes {
+		numInstances++
+		fmt.Println(fmt.Sprintf("-- Volume: %v", volume.Name))
```
sean-: Ha!  Don't we want `%s` here instead?  Or is `Name` something that doesn't satisfy the Stringer interface?

compute/volumes.go
```diff
@@ -0,0 +1,216 @@
...
+	State string
+	Type  string
+}
+
+func (c *VolumesClient) List(ctx context.Context, input *ListVolumesInput) ([]*Volume, error) {
+	fullPath := path.Join("/", c.client.AccountName, "volumes")
+
+	query := &url.Values{}
+	if input.Name != "" {
+		query.Set("name", input.Name)
```
sean-: This isn't critical for this PR, but `"name"`, `"size"`, `"state"`, and `"type"` should all be moved into a `const` block in the package so that we can find all call sites where a given parameter is used.  I'd probably use an unexported `const` identifier, too.  Something like:

```go
const (
  _VolumeName = "name"
)
```

If this were internal only and the inputs weren't coming from the caller, I'd suggest something like:

```go
type VolumeAttr string

const (
  _VolumeName _VolumeAttr = "name"
)
```

The only reason I'm saying that is `"name"` is a string with no type and isn't bound specifically to the Volume API and is a generic value that could be reused or hard to track down in the future.  Not critical, but something I've found that makes navigating code a bit easier, albeit more tedious to write up front.  Because this is at a customer-facing boundary, I'd only do the former, not the latter.

The extra commentary is only here for future code anthropologists.

compute/volumes.go
```diff
@@ -0,0 +1,216 @@
...
+	if input.Size != 0 {
+		result["size"] = input.Size
+	}
+
+	if input.Type != "" {
+		result["type"] = input.Type
+	}
+
+	if len(input.Networks) > 0 {
+		result["networks"] = input.Networks
```
sean-: Move to a constant.

examples/compute/volumes/list_volumes/main.go
```diff
@@ -0,0 +1,101 @@
...
+			Username:           userName,
+		}
+		signer, err = authentication.NewPrivateKeySigner(input)
+		if err != nil {
+			log.Fatalf("Error Creating SSH Private Key Signer: %v", err)
+		}
+	}
+
+	config := &triton.ClientConfig{
+		TritonURL:   os.Getenv("SDC_URL"),
```
sean-: `s/SDC_URL/TRITON_URL/`

compute/volumes_test.go
```diff
@@ -0,0 +1,451 @@
...
+	t.Run("eof", func(t *testing.T) {
+		testutils.RegisterResponder("GET", path.Join("/", accountURL, "volumes", fakeVolumeID), getVolumeEmpty)
+
+		_, err := do(context.Background(), computeClient)
+		if err == nil {
+			t.Fatal(err)
+		}
+
+		if !strings.Contains(err.Error(), "EOF") {
+			t.Errorf("expected error to contain EOF: found %s", err)
```
sean-: In the future make a sweep for `%s` as the verb for error messages and swap to `%v` that way we don't cargo cult `%s`.

sean-:
@stack72 Merged! Can you review the above comments in the PR and address separately?  Similarly, can you add the appropriate `CHANGELOG` entry?  Because this feature is whitelist enabled only, I'm not too worried about the missing `CHANGELOG` for a day or two.  :)

PR #107, opened 2018-01-19
Use proper query parameters to fix listing packages

cheapRoc:
This fixes an issue where we can't filter packages through CloudAPI itself. We weren't using query parameters where we should be.

~~This is still in play, I need to change the tests and add an acceptance test so we identify similar issues in the future.~~

### DIFF
--- a/compute/packages_test.go
+++ b/compute/packages_test.go
@@ -11,19 +11,111 @@ package compute_test
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"path"
 	"strings"
 	"testing"
 
+	"github.com/abdullin/seq"
+	triton "github.com/joyent/triton-go"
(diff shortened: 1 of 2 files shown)


compute/packages_test.go
```diff
@@ -11,12 +11,15 @@ package compute_test
...
 	"context"
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"path"
 	"strings"
 	"testing"
 
+	"github.com/abdullin/seq"
```
sean-: Where did you find this?  This looks like a very useful testing library.

Can you `dep ensure` and add `seq` to this PR?

cheapRoc: I can't take the credit, it came from James and the image API tests. But yes, it is handy.

cheapRoc:
Thanks @sean-, I need to run this by @stack72 tomorrow in terms of how he wants to handle acceptance testing then we can merge this in.

PR #109, opened 2018-01-24
Refactor current comments to make them mode idiomatic.

kwilczynski:
This commit changes comments for some of the exported variables and functions to
make them more idiomatic, and also adds comments where such were missing.

Also, remove the else statement from the `UserAgent()` function which is not
needed given that this function returns values immediately without doing
anything else.

### DIFF
--- a/version.go
+++ b/version.go
@@ -13,20 +13,30 @@ import (
 	"runtime"
 )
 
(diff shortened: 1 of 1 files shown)


stack72:
LGTM thanks for this @kwilczynski

cheapRoc:
👍

Issue #106, opened 2018-01-19
Add ChangeUserPassword support


cheapRoc:
Ref: #25

Add support for the `ChangeUserPassword` endpoint in CloudAPI.

cheapRoc:
If any of these remaining features aren't something we want to support, we can discuss here and/or delete the issues individually.

stack72:
Closed via #111

PR #111, opened 2018-01-28
Add support for the ChangeUserPassword call available in the CloudAPI.

kwilczynski:
This commit adds support for the ChangeUserPassword call available in the
CloudAPI, which allows for password to be updated for any valid Trition user.

### DIFF
--- a/identity/users.go
+++ b/identity/users.go
@@ -16,7 +16,8 @@ import (
 	"time"
 
 	"github.com/joyent/triton-go/client"
-	"github.com/pkg/errors"
+	"github.com/joyent/triton-go/errors"
+	pkgerrors "github.com/pkg/errors"
 )
 
 type UsersClient struct {
@@ -55,13 +56,13 @@ func (c *UsersClient) List(ctx context.Context, _ *ListUsersInput) ([]*User, err
 		defer respReader.Close()
 	}
 	if err != nil {
-		return nil, errors.Wrap(err, "unable to list users")
+		return nil, pkgerrors.Wrap(err, "unable to list users")
 	}
 
 	var result []*User
 	decoder := json.NewDecoder(respReader)
 	if err = decoder.Decode(&result); err != nil {
-		return nil, errors.Wrap(err, "unable to decode list users response")
+		return nil, pkgerrors.Wrap(err, "unable to decode list users response")
 	}
 
 	return result, nil
@@ -82,13 +83,13 @@ func (c *UsersClient) Get(ctx context.Context, input *GetUserInput) (*User, erro
 		defer respReader.Close()
 	}
 	if err != nil {
-		return nil, errors.Wrap(err, "unable to get user")
+		return nil, pkgerrors.Wrap(err, "unable to get user")
 	}
 
 	var result *User
 	decoder := json.NewDecoder(respReader)
 	if err = decoder.Decode(&result); err != nil {
-		return nil, errors.Wrap(err, "unable to decode get user response")
+		return nil, pkgerrors.Wrap(err, "unable to decode get user response")
 	}
 
 	return result, nil
@@ -109,7 +110,7 @@ func (c *UsersClient) Delete(ctx context.Context, input *DeleteUserInput) error 
 		defer respReader.Close()
 	}
 	if err != nil {
-		return errors.Wrap(err, "unable to delete user")
+		return pkgerrors.Wrap(err, "unable to delete user")
 	}
 
 	return nil
@@ -142,13 +143,13 @@ func (c *UsersClient) Create(ctx context.Context, input *CreateUserInput) (*User
 		defer respReader.Close()
 	}
 	if err != nil {
-		return nil, errors.Wrap(err, "unable to create user")
+		return nil, pkgerrors.Wrap(err, "unable to create user")
 	}
 
 	var result *User
 	decoder := json.NewDecoder(respReader)
 	if err = decoder.Decode(&result); err != nil {
-		return nil, errors.Wrap(err, "unable to decode create user response")
+		return nil, pkgerrors.Wrap(err, "unable to decode create user response")
 	}
 
 	return result, nil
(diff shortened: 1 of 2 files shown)


kwilczynski:
To do:

- [x] Implementation
- [x] Add unit tests
- [x] Add integration tests

kwilczynski:
I had an idea here, but I am not sure how closely we want to stick to a vanilla behaviour of the CloudAPI: we could check for equality between `password` and `password_confirmation` ourselves in the function, rather than having to pass this over to CloudAPI, and thus reduce/speed-up the feedback cycle for the user. In a case of strings being different, we could emit the same API error what CloudAPI would do.

The reason why this came through my mind, is primarily because I am not certain if an API for password changes/updates should change "password confirmation" at all, as I would imagine it should be something that one would offload to a CLI, etc., to deal with.

kwilczynski:
Closes https://github.com/joyent/triton-go/issues/106.

kwilczynski:
Unit tests:

```bash
$ go test ./identity/
ok  	github.com/joyent/triton-go/identity	0.021s
```

kwilczynski:
Acceptance tests:

```bash
$ TRITON_TEST=1 go test -v -run "TestAccUser_ChangeUserPassword" ./identity/
=== RUN   TestAccUser_ChangeUserPassword
2018/01/30 01:17:20 [INFO] Creating Triton Client with SSH Key Signer...
--- PASS: TestAccUser_ChangeUserPassword (9.74s)
PASS
ok  	github.com/joyent/triton-go/identity	9.784s
```

identity/users_test.go
```diff
@@ -39,6 +42,111 @@ func MockIdentityClient() *identity.IdentityClient {
...
 }
 
+func TestAccUser_ChangeUserPassword(t *testing.T) {
+	userLogin := testutils.RandPrefixString("TestAccUser", 32)
+	userEmail := fmt.Sprintf("%s@example.com", userLogin)
+	userPassword := testutils.RandString(32)
+	userNewPassword := testutils.RandString(32)
+
+	// Holds newly created user.
+	var newUser *identity.User
```
cheapRoc: The purpose of the state bag construct is to remove global state from the acceptance testing process. I'll discuss more in a bit.

identity/users_test.go
```diff
@@ -39,6 +42,111 @@ func MockIdentityClient() *identity.IdentityClient {
...
+				CallFunc: func(client interface{}) (interface{}, error) {
+					c := client.(*identity.IdentityClient)
+					ctx := context.Background()
+					input := &identity.CreateUserInput{
+						Email:    userEmail,
+						Login:    userLogin,
+						Password: userPassword,
+					}
+					user, err := c.Users().Create(ctx, input)
+					newUser = user
```
cheapRoc: Remove this and utilize the `stateBag` argument within your `CleanupFunc` implementation. More below...

identity/users_test.go
```diff
@@ -39,6 +42,111 @@ func MockIdentityClient() *identity.IdentityClient {
...
+					return user, err
+				},
+				CleanupFunc: func(client interface{}, _ interface{}) {
+					c := client.(*identity.IdentityClient)
+					ctx := context.Background()
+					input := &identity.DeleteUserInput{
+						UserID: newUser.ID,
```
cheapRoc: You want to leverage the `stateBag` which is the second `interface{}` argument in this function signature (currently named `_`). Using a type assertion you should receive a `identity.User` as this value.

I believe you want something like the following with a known test user name (`const` or randomly generated) for easier sweeping.

```go
CleanupFunc: func(client interface{}, stateBag interface{}) {
	user, userOk := stateBag.(*identity.User)
	if !userOk {
		log.Println("expected state to be User")
		return
	}

	if user.Name != testUserName {
		log.Printf("expected user to be named %s: found %s\n",
			testUserName, user.Name)
		return
	}

	c := client.(*identity.IdentityClient)
	ctx := context.Background()
	input := &identity.DeleteUserInput{
		UserID: user.ID,
	}
	err := c.Users().Delete(ctx, input)
	if err != nil {
		log.Println("could not delete user", user.ID)
	}
},
```

identity/users.go
```diff
@@ -181,13 +182,75 @@ func (c *UsersClient) Update(ctx context.Context, input *UpdateUserInput) (*User
...
+
+	return result, nil
+}
+
+// ChangeUserPasswordInput represents the options that can be specified
+// when changing an existing user password.
+type ChangeUserPasswordInput struct {
+	// The unique ID of the user for which to change the password. Required.
+	UserID string
+
```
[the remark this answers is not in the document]
kwilczynski: Will do. I simply followed style from some existing code. :)

identity/users_test.go
kwilczynski: True, and I really like the idea.

I only added this in a bid to pass user ID to an invocation of `StepAPICall` where I call `ChangeUserPassword`. I've noticed that it does not have support for passing the state bag (like the others), and using variable from more top-level scope in the closure was an option to retrieve the ID of newly created user.

identity/users_test.go
kwilczynski: Will do.

I decided to leverage the variable I had already (see comment above), but I state bag is the way to go, indeed.

Changes coming right up! :)

will: whyyyyyyyyyy did you put an @ before my name? it doesn't even make any sense

kwilczynski:
Hi @will, sorry to have accidentally pulled you into the code review! I was an accidental typo I corrected. :)

kwilczynski:
Hi @cheapRoc, I have made requested changes, and both unit and acceptance tests are passing:

(unit)

```bash
$ go test ./identity/
ok  	github.com/joyent/triton-go/identity	0.032s
```

(acceptance)
```bash
$ TRITON_TEST=1 go test -v -run "TestAccUser_ChangeUserPassword" ./identity/
=== RUN   TestAccUser_ChangeUserPassword
2018/01/30 22:33:11 [INFO] Creating Triton Client with SSH Key Signer...
--- PASS: TestAccUser_ChangeUserPassword (8.74s)
PASS
ok  	github.com/joyent/triton-go/identity	8.777s
```
Few question to you:

1) Would it make sense to add an ability to use the state bag to `CallFunc` from `testutils.StepAPICall`, and an ability to return an error (and thus halt) from the `CleanupFunc`?

2) I was also thinking, whether it would make sense to add support for checking/matching an error message to `testutils.StepAssertTritonError` alongside the `Code`, since some API calls return the same code but the error message can be different.

Also, sincere apologies for accidentally adding William to the code review in a typo-driven accident. :)

stack72:
This looks good to me now - thanks for all the work here @kwilczynski :) I am going to fix up a few typos locally!

Issue #105, opened 2018-01-19
Add SetRoleTags support


cheapRoc:
Ref: #25 

Add support for `SetRoleTags` endpoint in CloudAPI.

stack72:
Closed via #112

Issue #11, opened 2017-05-04
Add appropriate validation for all input structures


jen20:
We should likely add a `InputValidator` interface that defines `Validate` and call if the input struct implements it in the generic request handling part of the library.

PR #112, opened 2018-02-04
Add support for the SetRoleTags call available in the CloudAPI.

kwilczynski:
This commit adds support for the SetRoleTags call available in the
CloudAPI, which allows for setting a single or multiple role tags
to use with the RBAC, either for an individual resource or for the
whole group of resources (e.g., all the current instances, etc.).

Additionally, this commit also adds GetRoleTags function, whilst not
provided by the CloudAPI, can be used to retrieve a list of currently
set role tags.

### DIFF
--- a/identity/roles.go
+++ b/identity/roles.go
@@ -13,6 +13,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"path"
+	"strings"
 
 	"github.com/joyent/triton-go/client"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


kwilczynski:
To do:

- [x] Implementation
- [ ] Add unit tests
- [ ] Add integration tests

kwilczynski:
Closes https://github.com/joyent/triton-go/issues/105.

identity/roles.go
```diff
@@ -189,3 +190,117 @@ func (c *RolesClient) Delete(ctx context.Context, input *DeleteRoleInput) error
 
 	return nil
 }
+
+//
+type RoleTags struct {
+	//
+	Name string `json:"name"`
+	//
+	RoleTag []string `json:"role-tag"`
```
kwilczynski: I wonder if this should be made plural? Looks awkward to have a list there, but singular member name.

kwilczynski:
I've noticed that role tags are often returned in different order, for example the tag (from GetRoleTags) vs response after setting them (from SetRoleTags), thus I wonder whether we should sort these (lexicographically) for users convenience? I support in the greater scheme of things it does not matter as much when user, for example, walks over the slice with a range, etc.

identity/roles.go
```diff
@@ -189,3 +190,117 @@ func (c *RolesClient) Delete(ctx context.Context, input *DeleteRoleInput) error
...
+	decoder := json.NewDecoder(respReader)
+	if err = decoder.Decode(&result); err != nil {
+		return nil, errors.Wrap(err, "unable to decode set role tags response")
+	}
+
+	return result, nil
+}
+
+//
+func validateResourceType(resourceType string) error {
```
kwilczynski: Would suffice for now, but needs to be changed after https://github.com/joyent/triton-go/issues/11 is closed.

identity/roles.go
kwilczynski: After talking off-line with @stack72, we've decided to go with a plural form here, despite what the CloudAPI has, as it makes more sense since it is a list of role tags.

identity/roles.go
```diff
@@ -189,3 +190,81 @@ func (c *RolesClient) Delete(ctx context.Context, input *DeleteRoleInput) error
 
 	return nil
 }
+
+// RoleTags
```
sean-: comment on exported type RoleTags should be of the form "RoleTags ..." (with optional leading article)

identity/roles.go
```diff
@@ -189,3 +190,81 @@ func (c *RolesClient) Delete(ctx context.Context, input *DeleteRoleInput) error
...
+type RoleTags struct {
+	// Path to the resource.
+	Name string `json:"name"`
+	// The list of role tags assigned to this resource.
+	RoleTags []string `json:"role-tag"`
+}
+
+// GetRoleTagsInput
```
sean-: comment on exported type GetRoleTagsInput should be of the form "GetRoleTagsInput ..." (with optional leading article)

identity/roles.go
```diff
@@ -189,3 +190,81 @@ func (c *RolesClient) Delete(ctx context.Context, input *DeleteRoleInput) error
...
+
+	result := &RoleTags{
+		Name:     fullPath,
+		RoleTags: strings.Split(roleTags, ","),
+	}
+
+	return result, nil
+}
+
+// SetRoleTagsInput
```
sean-: comment on exported type SetRoleTagsInput should be of the form "SetRoleTagsInput ..." (with optional leading article)

identity/roles.go
kwilczynski: All the comments are coming right up! Thank you! 👍

identity/roles.go
kwilczynski: I decided to remove this and let the CloudAPI bubble an error should the resource type be different to the one supported. This removes the need to maintain any list of currently supported resources in the client.

kwilczynski:
Unit tests:

```bash
$ go test -v -run 'Test(Set|Get)RoleTags' ./identity/...
=== RUN   TestSetRoleTags
=== RUN   TestSetRoleTags/successful
=== RUN   TestSetRoleTags/error
=== RUN   TestSetRoleTags/eof
=== RUN   TestSetRoleTags/bad_decode
--- PASS: TestSetRoleTags (0.00s)
    --- PASS: TestSetRoleTags/successful (0.00s)
    --- PASS: TestSetRoleTags/error (0.00s)
    --- PASS: TestSetRoleTags/eof (0.00s)
    --- PASS: TestSetRoleTags/bad_decode (0.00s)
=== RUN   TestGetRoleTags
=== RUN   TestGetRoleTags/successful
=== RUN   TestGetRoleTags/error
--- PASS: TestGetRoleTags (0.00s)
    --- PASS: TestGetRoleTags/successful (0.00s)
    --- PASS: TestGetRoleTags/error (0.00s)
PASS
ok  	github.com/joyent/triton-go/identity	0.015s
```

kwilczynski:
Acceptance tests:

```bash
$ TRITON_TEST=1 go test -v -run "TestAccIdentity_" ./identity/
=== RUN   TestAccIdentity_SetRoleTags
2018/03/20 17:27:06 [INFO] Creating Triton Client with SSH Key Signer...
--- PASS: TestAccIdentity_SetRoleTags (8.68s)
=== RUN   TestAccIdentity_GetRoleTags
2018/03/20 17:27:15 [INFO] Creating Triton Client with SSH Key Signer...
--- PASS: TestAccIdentity_GetRoleTags (14.23s)
PASS
ok  	github.com/joyent/triton-go/identity	22.936s
```

identity/roles_test.go
```diff
@@ -11,38 +11,291 @@ package identity_test
...
+
+					input := &identity.CreateRoleInput{
+						Name: testRoleName,
+					}
+					role, err := c.Roles().Create(ctx, input)
+					newRole = role
+
+					// Allow a few seconds for the new role
+					// to propagate within CloudAPI.
+					time.Sleep(5 * time.Second)
```
kwilczynski: I don't have a better idea, sadly. I did try with a retry, but while polling `Roles().Get()` works fine, then setting tags through `SetRoleTags()` would fail anyway. Unless we wait, a bit... This is driving me crazy a bit :(

stack72: yeah such is eventual consistency!

stack72:
LGTM! Thanks for this :) Another one closer to cloudapi coverage

Issue #12, opened 2017-05-04
Add pagination via input structures where CloudAPI allows


cheapRoc:
Specifically ref: [https://apidocs.joyent.com/cloudapi/#ListMachines]()

It would be nice to have pagination query-like constructs where CloudAPI API requests allow it.

There's a note about this for ListMachines...

> You can paginate this API by passing in offset and limit. HTTP responses will contain the 
> additional headers x-resource-count and x-query-limit. If x-resource-count is less than 
> x-query-limit, you're done, otherwise call the API again with offset set to offset + limit to
> fetch additional instances.

kwilczynski:
The Node.js client uses the following library to handle pagination: https://github.com/joyent/node-lomstream.

PR #113, opened 2018-02-05
[WIP] Add support for getting a count of current instances from the CloudAPI.

kwilczynski:
This commit adds a `Count` method which can be used to get a total
number of currently provisioned instances as a single integer value,
and does not retrieve any details of each of the instances.

### DIFF
--- a/compute/instances.go
+++ b/compute/instances.go
@@ -16,6 +16,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


kwilczynski:
To do:

- [x] Implementation
- [ ] Add unit tests
- [ ] Add integration tests

kwilczynski:
Related to https://github.com/joyent/triton-go/issues/12.

kwilczynski:
The following other resource types seem to also support `HEAD` request to get the total count via the `X-Resource-Count` header: Machines (this Pull Request), Networks, NICs, and Volumes; as per what I could find in the https://github.com/joyent/sdc-cloudapi repository (looking at the backend implementation).

kwilczynski:
To add :-

The inspiration to add this came while I looked at #12, and then looked through the https://github.com/joyent/sdc-cloudapi repository.

compute/instances.go
```diff
@@ -103,6 +104,39 @@ func (gmi *GetInstanceInput) Validate() error {
 	return nil
 }
 
+//
+type CountInstancesInput struct{}
```
stack72: is this even worth adding? There are some other input structs in other packages but those are areas in which filters can occur

compute/instances.go
```diff
@@ -103,6 +104,39 @@ func (gmi *GetInstanceInput) Validate() error {
...
+
+//
+func (c *InstancesClient) Count(ctx context.Context, _ *CountInstancesInput) (int, error) {
+	fullPath := path.Join("/", c.client.AccountName, "machines")
+
+	reqInputs := client.RequestInput{
+		Method: http.MethodHead,
```
stack72: +1 for the use of HEAD request here

compute/instances.go
```diff
@@ -103,6 +104,39 @@ func (gmi *GetInstanceInput) Validate() error {
...
+		Path:   fullPath,
+	}
+
+	response, err := c.client.ExecuteRequestRaw(ctx, reqInputs)
```
stack72: I think you need to check `err != nil` here

kwilczynski: Well spotted! Fixed.

compute/instances.go
kwilczynski: I asked myself the same question "should there be an input parameter here at all?", then I looked at some other examples e.g. https://github.com/joyent/triton-go/blob/master/account/keys.go#L41 (and few others with similar concept) and decided to follow the canon.

Should we remove it?

stack72: on second thoughts, maybe it's best to leave it for now - it may allow us to count instances based on a filter later :)

kwilczynski: I am happy to remove it. Since we are using `HEAD` against `/:account/machines`, there is no way to filter on anything - as there is no body aside of headers, and I use `X-Resource-Count` header to get the machines count back. :)

stack72: ok let's remove it then :)

kwilczynski: Done!

stack72:
Closed by #119

PR #116, opened 2018-02-12
compute/instances: List instances now adds support for querying by tag

stack72:


kwilczynski:
Looks good to me! 👍 

Definitely addresses the lack of support for tags, and build a groundwork for the future support of the `Tag` type. ❤️

PR #119, opened 2018-02-19
Add support for getting a count of current instances from the CloudAPI.

stack72:
This commit adds a `Count` method which can be used to get a total
number of currently provisioned instances as a single integer value,
and does not retrieve any details of each of the instances.

This superceeds #113

### DIFF
--- a/compute/instances.go
+++ b/compute/instances.go
@@ -16,6 +16,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"strconv"
 	"strings"
 	"time"
 
@@ -103,6 +104,38 @@ func (gmi *GetInstanceInput) Validate() error {
 	return nil
 }
 
+func (c *InstancesClient) Count(ctx context.Context, input *ListInstancesInput) (int, error) {
+	fullPath := path.Join("/", c.client.AccountName, "machines")
+
+	reqInputs := client.RequestInput{
+		Method: http.MethodHead,
+		Path:   fullPath,
+		Query:  buildQueryFilter(input),
+	}
+
+	response, err := c.client.ExecuteRequestRaw(ctx, reqInputs)
+	if err != nil {
+		return -1, pkgerrors.Wrap(err, "unable to get machines count")
+	}
+
+	if response == nil {
+		return -1, pkgerrors.New("request to get machines count has empty response")
+	}
+	defer response.Body.Close()
+
+	var result int
+
+	if count := response.Header.Get("X-Resource-Count"); count != "" {
+		value, err := strconv.Atoi(count)
+		if err != nil {
+			return -1, pkgerrors.Wrap(err, "unable to decode machines count response")
+		}
+		result = value
+	}
+
+	return result, nil
+}
+
 func (c *InstancesClient) Get(ctx context.Context, input *GetInstanceInput) (*Instance, error) {
 	if err := input.Validate(); err != nil {
 		return nil, pkgerrors.Wrap(err, "unable to get machine")
@@ -159,9 +192,7 @@ type ListInstancesInput struct {
 	Credentials bool
 }
 
-func (c *InstancesClient) List(ctx context.Context, input *ListInstancesInput) ([]*Instance, error) {
-	fullPath := path.Join("/", c.client.AccountName, "machines")
-
+func buildQueryFilter(input *ListInstancesInput) *url.Values {
 	query := &url.Values{}
 	if input.Brand != "" {
 		query.Set("brand", input.Brand)
@@ -175,10 +206,10 @@ func (c *InstancesClient) List(ctx context.Context, input *ListInstancesInput) (
 	if input.State != "" {
 		query.Set("state", input.State)
 	}
-	if input.Memory >= 1 && input.Memory <= 1000 {
+	if input.Memory >= 1 {
 		query.Set("memory", fmt.Sprintf("%d", input.Memory))
 	}
-	if input.Limit >= 1 {
+	if input.Limit >= 1 && input.Limit <= 1000 {
 		query.Set("limit", fmt.Sprintf("%d", input.Limit))
 	}
 	if input.Offset >= 0 {
@@ -199,10 +230,16 @@ func (c *InstancesClient) List(ctx context.Context, input *ListInstancesInput) (
 		}
 	}
 
+	return query
+}
+
+func (c *InstancesClient) List(ctx context.Context, input *ListInstancesInput) ([]*Instance, error) {
+	fullPath := path.Join("/", c.client.AccountName, "machines")
+
 	reqInputs := client.RequestInput{
 		Method: http.MethodGet,
 		Path:   fullPath,
-		Query:  query,
+		Query:  buildQueryFilter(input),
 	}
 	respReader, err := c.client.ExecuteRequest(ctx, reqInputs)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


compute/instances.go
```diff
@@ -159,9 +192,7 @@ type ListInstancesInput struct {
 	Credentials bool
 }
 
-func (c *InstancesClient) List(ctx context.Context, input *ListInstancesInput) ([]*Instance, error) {
-	fullPath := path.Join("/", c.client.AccountName, "machines")
-
+func buildQueryFilter(input *ListInstancesInput) *url.Values {
```
kwilczynski: Very nice idea 👍

kwilczynski:
I tested sending `HEAD` against the CloudAPI, and it seem to work as intended, as per:

(with filter)

```sh
krzysztof.wilczynski@deupc00156 ~/.../joyent/triton-go (master $)$ cloudapi_head '/kwilczynski/machines?docker=false'
HTTP/1.1 200 OK
Server: cloudapi/8.3.0
Request-Id: 1675a22d-8f55-46d6-a0ce-fa2981dc4bcd
x-query-limit: 1000
role-tag: test,test123
Content-Type: application/json
Access-Control-Expose-Headers: Api-Version, Request-Id, Response-Time
Strict-Transport-Security: max-age=6570000
Date: Mon, 19 Feb 2018 19:52:24 GMT
Content-MD5: 11FxOYiYfpMxmANj4kGJzg==
Access-Control-Allow-Origin: *
Response-Time: 667
Connection: Keep-Alive
x-resource-count: 0
Access-Control-Allow-Headers: Accept, Accept-Version, Content-Length, Content-MD5, Content-Type, Date, Api-Version, Response-Time
Access-Control-Allow-Methods: POST, GET, HEAD
Api-Version: 8.0.0
Content-Length: 2
```

(without filter)

```sh
krzysztof.wilczynski@deupc00156 ~/.../joyent/triton-go (master $)$ cloudapi_head '/kwilczynski/machines'
HTTP/1.1 200 OK
Server: cloudapi/8.3.0
Request-Id: f78be80d-8708-485c-9143-d1eb5a183518
x-query-limit: 1000
role-tag: test,test123
Content-Type: application/json
Access-Control-Expose-Headers: Api-Version, Request-Id, Response-Time
Strict-Transport-Security: max-age=6570000
Date: Mon, 19 Feb 2018 19:52:31 GMT
Content-MD5: ECyZsF+Ia6bpVAciJcXQuQ==
Access-Control-Allow-Origin: *
Response-Time: 1161
Connection: Keep-Alive
x-resource-count: 1
Access-Control-Allow-Headers: Accept, Accept-Version, Content-Length, Content-MD5, Content-Type, Date, Api-Version, Response-Time
Access-Control-Allow-Methods: POST, GET, HEAD
Api-Version: 8.0.0
Content-Length: 692
```

kwilczynski:
There is another thing that I meant to address in an issue, but we might as well try to tackle this now.

The following (and similar ones):

https://github.com/joyent/triton-go/blob/8b12b82dc02a9395ca1506bc6a8ed0dbfc46977a/compute/instances.go#L190-L192

Are potentially an issue, but why? Because of the Go zero-values and the way how we set the parameter here. Allow me to explain:

The CloudAPI allows one to set the something along these lines: `docker=false` or `docker=true`; as a filter. Which is a perfectly valid use case, since one might wish to filter out container (Docker?) instances/machines from the results, but keep everything else.

At the moment, this is not possible. Why? Because of the `input.Docker` being a `boolean` type, which has its zero-value set to `false` by default. This means that the following happens:

**1) User sets `Docker: false` in the input type;**

This will result in no `docker=false` being added as a query parameter in the entire query string, and as per the default in the CloudAPI (which is akin to setting up `docker=true`), all the instances/machines would be returned regardless of the user intent here.

**2) User sets `Docker: true` in the input type;**

This will result in `docker=true` being added as a query parameter to the query string explicitly. This is also the same as the CloudAPI default.

**3) User does not provide neither `Docker: false` nor `Docker: true` in the input type;**

The Go zero-value will take into effect here, which is `false` by default for the `boolean` type. This will result in CloudAPI following its default behaviour, which is to include container instances/machines in the output.

We need to fix this, if we want to support filter as a "toggle" allowing for a combination of different values.

kwilczynski:
Apologies for being a bit of a drag on the Pull Request, but I have another question. The following:

https://github.com/joyent/triton-go/blob/8b12b82dc02a9395ca1506bc6a8ed0dbfc46977a/compute/instances.go#L178

I believe, that checking for >= for the memory would be fine, but I would not put a top-level ceiling here, as we might introduce instances with a TiB or more memory in the future (premium?).

Where this check would make more sense, would be the following:

https://github.com/joyent/triton-go/blob/8b12b82dc02a9395ca1506bc6a8ed0dbfc46977a/compute/instances.go#L181

Since the limit can only be as high as 1000, which is its top-level limit.

Albeit, this is just a question, as I would need to confirm what happens when one selects more than a value of 1000 for the memory filter.

kwilczynski:
To add to the https://github.com/joyent/triton-go/pull/119#issuecomment-366793077.

I think this is definitely achievable by having these fields as pointers, so that we could dereference the pointer to get the actual stored value should it be a non-nil value. But that is getting us closer to what AWS SDK does almost everywhere and it is definitely error prone (on the user side, most of the time), and might require helper methods (e.g., `aws.String(s string)`) to be provided.

compute/instances.go
```diff
@@ -175,7 +206,7 @@ func (c *InstancesClient) List(ctx context.Context, input *ListInstancesInput) (
 	if input.State != "" {
 		query.Set("state", input.State)
 	}
-	if input.Memory >= 1 && input.Memory <= 1000 {
+	if input.Memory >= 1 {
 		query.Set("memory", fmt.Sprintf("%d", input.Memory))
 	}
 	if input.Limit >= 1 {
```
kwilczynski: Limit greater than 1000 results in _409 Conflict_.

compute/instances.go
```diff
@@ -103,6 +104,38 @@ func (gmi *GetInstanceInput) Validate() error {
 	return nil
 }
 
+func (c *InstancesClient) Count(ctx context.Context, input *ListInstancesInput) (int, error) {
```
kwilczynski: Here, the `ListInstancesInput` might be an interesting use-case for much dreaded type aliases added in Go 1.9. So that would be:

```go
type CountInstancesInput = ListInstancesInput
```

See https://golang.org/doc/go1.9#language for more details.

kwilczynski: Example: https://play.golang.org/p/ftoqWLjx154

compute/instances.go
kwilczynski: Example:

```sh
$ for t in memory limit; do \
>     for v in 0 1 1000 10000; do \
>         echo -ne "${t}=${v}:\t"; cloudapi_head "/kwilczynski/machines?${t}=${v}" 2>/dev/null | head -1;
>     done
> done
memory=0:	HTTP/1.1 200 OK
memory=1:	HTTP/1.1 200 OK
memory=1000:	HTTP/1.1 200 OK
memory=10000:	HTTP/1.1 200 OK
limit=0:	HTTP/1.1 409 Conflict
limit=1:	HTTP/1.1 200 OK
limit=1000:	HTTP/1.1 200 OK
limit=10000:	HTTP/1.1 409 Conflict
```

kwilczynski:
Hi @stack72, looks good to me! Let's merge!

stack72:
Thanks for the review here @kwilczynski :)

PR #121, opened 2018-02-20
client/ExecuteRequestRaw: Add support for querystring in this func

stack72:


kwilczynski:
Hi @stack72, looks good! A very nice feature that was missing.

PR #126, opened 2018-02-23
root: Move GetTritonEnv into root triton package

cheapRoc:
Something I noticed while working on `triton-go` lately is that we should probably rename `client.GetTritonEnv` => `triton.GetEnv`. This will help maintain our environment variable configuration at the root of the project for easier consumption. I've also added a unit test.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -18,7 +18,6 @@ import (
 	"net"
 	"net/http"
 	"net/url"
-	"os"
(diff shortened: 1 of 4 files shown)


stack72:
Nice addition @cheapRoc I have added similar code elsewhere due to a circular dep so I can remove that now :)

PR #127, opened 2018-02-24
Alpha TSG client

cheapRoc:
At the moment this is required to demo and work on authentication within TSG. I don't believe this is something we want to merge straight into master but there are a fix or two that I can peel off for logic already in master.

The goal was to present something that would hit TSG's templates endpoint and pass along a signed HTTP request we can re-purpose for authenticating a user within TSG on behalf of CloudAPI.

WIP ~ DO NOT MERGE

### DIFF
--- a/services/templates.go
+++ b/services/templates.go
@@ -64,11 +64,11 @@ func (c *TemplatesClient) List(ctx context.Context, _ *ListTemplatesInput) ([]*I
 }
 
 type GetTemplateInput struct {
-	ID int64
+	ID string
 }
 
 func (i *GetTemplateInput) Validate() error {
-	if i.ID == 0 {
+	if i.ID == "" {
 		return fmt.Errorf("template id can not be empty")
 	}
 
@@ -80,10 +80,9 @@ func (c *TemplatesClient) Get(ctx context.Context, input *GetTemplateInput) (*In
 		return nil, pkgerrors.Wrap(err, "unable to get instance template")
 	}
 
-	fullPath := path.Join(templatesPath, fmt.Sprintf("%d", input.ID))
 	reqInputs := client.RequestInput{
 		Method: http.MethodGet,
-		Path:   fullPath,
+		Path:   path.Join(templatesPath, input.ID),
 	}
 	respReader, err := c.client.ExecuteRequestTSG(ctx, reqInputs)
 	if respReader != nil {
@@ -178,11 +177,11 @@ func (c *TemplatesClient) Create(ctx context.Context, input *CreateTemplateInput
 }
 
 type DeleteTemplateInput struct {
-	ID int64
+	ID string
 }
 
 func (i *DeleteTemplateInput) Validate() error {
-	if i.ID == 0 {
+	if i.ID == "" {
 		return fmt.Errorf("template name can not be empty")
 	}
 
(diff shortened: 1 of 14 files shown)


cheapRoc:
I've opened #130 from off thread discussion around TSG endpoint configuration.

services/templates.go
```diff
@@ -0,0 +1,213 @@
...
+
+package services
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"path"
+
```
sean-: struct field AccountId should be AccountID

services/templates.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/joyent/triton-go/client"
+	pkgerrors "github.com/pkg/errors"
```
sean-: struct field ImageId should be ImageID

services/templates.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	fullPath := path.Join("/v1/tsg/templates", input.Name)
+	reqInputs := client.RequestInput{
+		Method: http.MethodGet,
+		Path:   fullPath,
+	}
+	respReader, err := c.client.ExecuteRequestTSG(ctx, reqInputs)
+	if respReader != nil {
+		defer respReader.Close()
```
sean-: struct field AccountId should be AccountID

services/templates.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	if err != nil {
```
sean-: struct field ImageId should be ImageID

cheapRoc:
For this to be useful it currently needs all open PRs under joyent/triton-service-groups as well as joyent/triton-service-groups#7 and joyent/triton-service-groups#8.

services/templates.go
cheapRoc: Yeah I know, we need to change `AccountId` within [joyent/triton-service-groups](http://github.com/joyent/triton-service-groups) first.

cheapRoc: Actually, with joyent/triton-service-groups#13 merged I can just use struct tags for the JSON parsing and rename to `AccountID` just on the `triton-go` side.

@stack72 opened joyent/triton-service-groups#14 for the renaming.

services/templates.go
```diff
@@ -0,0 +1,205 @@
...
+	"net/http"
+	"path"
+
+	"github.com/joyent/triton-go/client"
+	pkgerrors "github.com/pkg/errors"
+)
+
+const templatesPath = "/v1/tsg/templates"
+
+type TemplatesClient struct {
```
sean-: struct field AccountId should be AccountID

services/templates.go
```diff
@@ -0,0 +1,205 @@
...
+	client *client.Client
+}
```
sean-: struct field ImageId should be ImageID

cheapRoc:
Ok, we're all good on renames and JSON tags.

cheapRoc:
I've rebased and removed the static services URL. There is now a manual override for local development (`TRITON_TSG_URL=http://localhost:3000/` env var). There's also URL generation based on the currently selected DC (`TRITON_URL`) for when the API goes live. This transforms `https?://(.*).api.joyent.com` into `https://tsg.%s.svc.joyent.zone`. This all means you can configure a `services` client by providing the same client configuration/authentication as you would today with `triton-go` (even if TSG doesn't include auth yet).

Examples have been updated under `examples/services/` and pair with latest TSG master ([3d7c120](https://github.com/joyent/triton-service-groups/commit/3d7c120eec951e9b11c44fc36a73821f68f95801)).

cheapRoc:
Going to close and re-open with the branch I'm actually using for TF integration.

PR #132, opened 2018-02-28
compute/instances: Add missing Close() for the response body.

kwilczynski:
This commit adds missing call to the `Close()` method for the underlying
`io.ReadCloser` interface, which an cause a potential memory leak.

### DIFF
--- a/compute/instances.go
+++ b/compute/instances.go
@@ -245,6 +245,9 @@ func (c *InstancesClient) List(ctx context.Context, input *ListInstancesInput) (
 		Query:  buildQueryFilter(input),
 	}
 	respReader, err := c.client.ExecuteRequest(ctx, reqInputs)
+	if respReader != nil {
+		defer respReader.Close()
+	}
 	if err != nil {
 		return nil, pkgerrors.Wrap(err, "unable to list machines")
 	}


kwilczynski:
**Note:** I went with the most defensive approach, and I am checking for whether the response object is not `nil` before calling `defer`.

kwilczynski:
There isn't an integration test for `compute.List()` (that will be added in a different Pull Request), so in the meantime I put a quick test to verify that everything is still fine, results:

```bash
$ go run main.go
([]*compute.Instance) (len=1 cap=1) {
 (*compute.Instance)(0xc420085500)({
  ID: (string) (len=36) "1575d90d-7d6a-6c6a-b690-99a4dac001ab",
  Name: (string) (len=4) "test",
  Type: (string) (len=12) "smartmachine",
  Brand: (string) (len=2) "lx",
  State: (string) (len=7) "running",
  Image: (string) (len=36) "b90a3f67-5b0b-d304-5f09-1b49d406a3ce",
  Memory: (int) 128,
  Disk: (int) 3072,
  Metadata: (map[string]string) {
  },
  Tags: (map[string]interface {}) (len=3) {
   (string) (len=21) "triton.network.public": (string) (len=17) "Joyent-SDC-Public",
   (string) (len=31) "docker:label:portal:transaction": (string) (len=8) "79651c08",
   (string) (len=10) "sdc_docker": (bool) true
  },
  Created: (time.Time) 2018-01-27 23:40:18.059 +0000 UTC,
  Updated: (time.Time) 2018-01-27 23:40:30 +0000 UTC,
  Docker: (bool) true,
  IPs: ([]string) (len=2 cap=4) {
   (string) (len=13) "192.168.128.5",
   (string) (len=11) "72.2.113.82"
  },
  Networks: ([]string) (len=2 cap=4) {
   (string) (len=36) "2797b6ab-0ef2-49ac-a359-0959bca25cad",
   (string) (len=36) "2bf2a968-a4da-414b-a997-b3a8ff92de0b"
  },
  PrimaryIP: (string) (len=11) "72.2.113.82",
  FirewallEnabled: (bool) true,
  ComputeNode: (string) (len=36) "44454c4c-3800-104c-8052-b6c04f383432",
  Package: (string) (len=15) "g4-highcpu-128M",
  DomainNames: ([]string) <nil>,
  CNS: (compute.InstanceCNS) {
   Disable: (bool) false,
   ReversePTR: (string) "",
   Services: ([]string) <nil>
  }
 })
}
```

PR #135, opened 2018-03-12
Manta RBAC authentication support.

stack72:
The RBAC HTTP Signatures between manta and triton are slightly different

Therefore, we need to generate the signature based on whether the request
is for RBAC AND for Manta or not

We therefore, moved the logic to a central generate func rather than being
handled different;y between private key and ssh agent signers

Paired on this with @cheapRoc

### DIFF
--- a/authentication/agent_key_identifier.go
+++ b/authentication/agent_key_identifier.go
@@ -1,0 +1,25 @@
+package authentication
+
+import "path"
+
+type KeyID struct {
+	UserName    string
+	AccountName string
+	Fingerprint string
+	IsManta     bool
+}
+
+func (input *KeyID) generate() string {
+	var keyID string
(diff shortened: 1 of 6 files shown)


kwilczynski:
For posterity:

- CloudAPI: https://apidocs.joyent.com/cloudapi/#appendix-c-http-signature-authentication
- Manta: https://apidocs.joyent.com/manta/api.html#authentication

cheapRoc:
@jen20 Can we get your blessing/thoughts since you've been working on HTTP sig auth lately.

jen20:
@cheapRoc, @stack72, this looks fine to fix up authentication short term, however i think ultimately we should use the `go-httpauthentication` library here - it supports a wider range of functionality from the standard and has been simplified quite dramatically.

PR #144, opened 2018-04-04
cli/account: Add the ability to manage Triton account from CLI

stack72:
This adds GET and UPDATE for triton account command

This PR also needs to change the Account Update request to use a
POST rather than a PUT as per the API documentation. Without this,
it doesn't actually make the changes in the API


account/account.go
```diff
@@ -81,7 +81,7 @@ type UpdateInput struct {
 func (c AccountClient) Update(ctx context.Context, input *UpdateInput) (*Account, error) {
 	fullPath := path.Join("/", c.Client.AccountName)
 	reqInputs := client.RequestInput{
-		Method: http.MethodPut,
+		Method: http.MethodPost,
```
kwilczynski: 👍 +1

PR #152, opened 2018-04-27
client: Not Setting Services URL on Private Install

stack72:
Anything other than JPC / SPC will no longer try and parse the
datacenter to set the services url


client/client_test.go
```diff
@@ -25,6 +25,7 @@ func TestNew(t *testing.T) {
 	spcTritonURL := "https://us-east-1.api.samsungcloud.io"
 	jpcServiceURL := "https://tsg.us-east-1.svc.joyent.zone"
 	spcServiceURL := "https://tsg.us-east-1.svc.samsungcloud.zone"
+	privateInstallUrl := "https://myinstall.mycompany.com"
```
golangci: var privateInstallUrl should be privateInstallURL

PR #159, opened 2018-12-19
Support Go Modules

SamWhited:
Hello,

Please consider supporting [Go Modules], the new packaging standard that replaces Dep and will be adopted fully in Go 1.12. Experimental support is in Go 1.11 and the new module paths are supported in Go 1.9.7+ and Go 1.10.3+ in a read-only manner for backwards compatibility with all supported versions of Go.

Because this library is still below version 2 the `go.mod` file is fairly simple. The only other thing that would need to happen is that future releases would need to have semver compatible tags that start with a "v" (unlike the current convention which does not include the "v" and is not supported, for example the next version should be `v1.3.2`)

Thank you for your consideration.

[Go Modules]: https://github.com/golang/go/wiki/Modules


SamWhited:
Gentle ping again.

SamWhited:
Another quick ping since a tag that's not valid according to Go modules was just made.

twhiteman:
Good point on the semver tags - I was wondering why we'd not used a 'vX.Y.Z' tag before (like most other go repos), seems it is just inertia. I've started adding 'vX.Y.Z' tags as of 'v1.6.0'.

The v1.7.0 version brings in support for go modules. In a later (e.g. v2.0.0) we can drop the vendor directories from the repo.

Thanks for the pull request (sorry I didn't merge your PR - but I wanted to work on that myself, and test to be sure the modules were working and up-to-date correctly before I committed).

PR #171, opened 2020-04-15
Add build tag to ignore dummy.go from builds

picatz:
👋Hello!

This PR introduces a build tag to exclude the dummy credentials from being included in built binaries using this package.

## Before

```console
$ file before
before: Mach-O 64-bit executable x86_64  
$ strings before | grep "RSA PRIVATE"
-----BEGIN RSA PRIVATE KEY-----
-----END RSA PRIVATE KEY-----
```

## After

```console
$ file after
after: Mach-O 64-bit executable x86_64
$ strings after | grep "RSA PRIVATE"
```

### DIFF
--- a/authentication/dummy.go
+++ b/authentication/dummy.go
@@ -1,3 +1,4 @@
+// +build ignore
 //
 // Copyright (c) 2018, Joyent, Inc. All rights reserved.
 //


twhiteman:
Thanks - that cleans that up!

PR #172, opened 2020-04-15
Remove unused versions from go.sum

dnephin:
Hello!

Every project which depends on `triton-go`, and uses go modules, will have to include the full list of versions in its own go.sum. This resulted in our `go.sum` having thousands of unnecessary lines.

I ran `go mod tidy` on `triton-go` which removed all of the unused versions and dependencies from both `go.mod` and `go.sum`.


twhiteman:
Thanks - that's much more manageable now :)

PR #176, opened 2020-04-24
Add support for volume tags

### LINKED ISSUE #175: Add support for volume tags
Volumes now support tags (labels), so we should support tags in volume create and volume list.

twhiteman:
Fixes #175.


twhiteman:
Yes, I only want it as strings - I believe it was a mistake to allow a mix of strings/bools/integers for vm.tags (so messy).

PR #185, opened 2020-05-06
Instance Metadata can take either String, Boolean or Numeric JSON values

kusor:


compute/instances.go
```diff
@@ -67,7 +67,7 @@ type Instance struct {
 	Image              string                 `json:"image"`
 	Memory             int                    `json:"memory"`
 	Disk               int                    `json:"disk"`
-	Metadata           map[string]string      `json:"metadata"`
+	Metadata           map[string]interface{} `json:"metadata"`
```
twhiteman, line 70: This will be a breaking change - so we'll need to bump the major version and make notes accordingly.

compute/instances.go
```diff
@@ -287,7 +287,7 @@ type CreateInstanceInput struct {
 	LocalityStrict  bool
 	LocalityNear    []string
 	LocalityFar     []string
-	Metadata        map[string]string
+	Metadata        map[string]interface{}
 	Tags            map[string]string //
```
twhiteman, line 291: Should this Tags input be allowed to take `interface{}` too?

kusor: VMAPI doesn't disallow it, so I'd say it should, yeah. I'll update.

compute/instances.go
kusor: Sounds good, I'll update.

go.mod
```diff
@@ -1,4 +1,6 @@
-module github.com/joyent/triton-go
+module github.com/joyent/triton-go/v2
+
+go 1.14
```
twhiteman, line 3: I don't think we want go 1.14 here (this means the minimum go version required... maybe?) - so perhaps just drop this line. I currently use 1.12 and everything builds/works fine (even with this go 1.14 line there).

go.mod
```diff
@@ -11,7 +13,8 @@ require (
 	github.com/mattn/go-isatty v0.0.8
 	github.com/mattn/go-runewidth v0.0.3 // indirect
 	github.com/olekukonko/tablewriter v0.0.0-20180130162743-b8a9be070da4
-	github.com/pkg/errors v0.8.1
+	github.com/pelletier/go-toml v1.7.0 // indirect
```
twhiteman, line 16: I don't think we require go-toml? Maybe double check by running `go mod tidy`

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
...
+**NOTE:** This is a precursor release of triton v2.0.0, including backwards
+incompatible changes, like the modification of Instance's Metadata and Tags
+fields from `map[string][string]` and optionally `map[string]interface{}` to
+`map[string]interface{}` everywhere. This change suits better with the possible
+types of Metadata and Tags as defined by Triton's VMAPI: valid strings,
+booleans or numeric JSON values.
+
+- Fix List Instances type error due to Instance Metadata including boolean
+or numeric values [#166].
+
```
twhiteman, line 14: Maybe make a note of code will need to reference using the `triton-go/v2` namespace?

Issue #197, opened 2021-03-11
Support pkcs8


Smithx10:
Currently,  newer openssh keys aren't parsed correctly when they are provided by TRITON_KEY_MATERIAL || SDC_KEY_MATERIAL.

PR #198, opened 2021-03-11
added pkcs8 and edcsa for private_key_signer

Smithx10:
Addresses issue #197

### DIFF
--- a/authentication/ssh_agent_signer.go
+++ b/authentication/ssh_agent_signer.go
@@ -73,7 +73,7 @@ func NewSSHAgentSigner(input SSHAgentSignerInput) (*SSHAgentSigner, error) {
 		return nil, err
 	}
 	signer.key = matchingKey
-        signer.formattedKeyFingerprint, err = formatPublicKeyFingerprint(signer.key, true)
+	signer.formattedKeyFingerprint, err = formatPublicKeyFingerprint(signer.key, true)
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to format match public key")
 	}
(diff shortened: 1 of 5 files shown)


authentication/ssh_agent_signer.go
```diff
@@ -72,7 +73,10 @@ func NewSSHAgentSigner(input SSHAgentSignerInput) (*SSHAgentSigner, error) {
 		return nil, err
 	}
 	signer.key = matchingKey
-	signer.formattedKeyFingerprint = formatPublicKeyFingerprint(signer.key, true)
+        signer.formattedKeyFingerprint, err = formatPublicKeyFingerprint(signer.key, true)
```
twhiteman, line 76: Indentation is weird here.

twhiteman (changes requested):
Question: This does not affect consumers of triton-go, correct? I.e. they don't need to change any code due to this change?

authentication/ssh_agent_signer.go
Smithx10: @twhiteman, 

Question: This does not affect consumers of triton-go, correct? I.e. they don't need to change any code due to this change?

Answer:  Correct.

This functionality just expands the type of private key material that authentication.NewPrivateKeySigner() can use.

Is an example of how it is used:
https://github.com/joyent/triton-go/blob/master/examples/compute/instances/main.go#L73

Currently you can see here that we were only able to parse PKCS1 Private Keys.  Any other keys required the use of the ssh-agent.
https://github.com/joyent/triton-go/blob/master/authentication/private_key_signer.go#L51

sdc-cloudapi supports DSA,RSA, and ECDSA. (https://github.com/joyent/sdc-cloudapi/blob/master/lib/auth.js#L63) but I'm hesitant to add DSA due to its fragility. https://github.com/golang/go/issues/40337

twhiteman (approved):

